PR #7, opened 2015-01-04
Minor typo fixes and add Archlinux support

brenix:
Found this amazing role and noticed a couple typos. Also wanted to add support for Archlinux as you've made it very easy to do! Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ sshd_ListenAddress:
 sshd_skip_defaults: true
 sshd:
   Compression: true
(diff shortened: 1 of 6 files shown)


mattwillsher:
Hi @brenix, thanks for the PR and the contained fixes and additions.

It all looks good to me.

There is a slight change to your PR - the ordering of the SSH option is controlled in meta/options_body, which is then used to regenerate the template. I'll make that change after the merge though as it doesn't impact the content.

Thanks again1

PR #14, opened 2015-05-19
Remove apt role dependancy

bensallen:
I'd like to remove the apt role dependency as ansible-playbook doesn't evaluate the when statement in meta/main.yml when checking for installed role dependancies, so the apt role must be installed even when not managing any apt systems. It doesn't appear this sshd roles requires any functionality in the apt role explicitly.

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -24,6 +24,4 @@ galaxy_info:
   categories:
   - networking
   - system
-dependencies:
-  - role: willshersystems.apt
-    when: ansible_pkg_mgr == 'apt' and apt_has_run is not defined
+dependencies: []


mattwillsher:
I'm sure that used to work... Which version of Ansible are you using?

bensallen:
Testing on a fresh RHEL7 virtual machine:

Testing with ansible 2.0.0 from the devel branch:

    [root@localhost ansible-sshd]# ansible-playbook --version
    ansible-playbook 2.0.0
      configured module search path = None
    [root@localhost ansible-sshd]# ansible-playbook -i tests/inventory tests/test.yml --connection=local
    ERROR: cannot find role in /root/ansible-sshd/tests/roles/willshersystems.apt or /root/ansible-sshd/tests/willshersystems.apt or /root/willshersystems.apt

Testing with Ansible 1.9.1 from EPEL:
    
    # ansible-playbook --version
    ansible-playbook 1.9.1
      configured module search path = None
    [root@localhost ansible-sshd]# ansible-playbook -i tests/inventory tests/test.yml --connection=local
    ERROR: cannot find role in /root/ansible-sshd/tests/roles/willshersystems.apt or /root/ansible-sshd/tests/willshersystems.apt or /root/willshersystems.apt

I'd think logically this makes sense from ansible's prospective as you may be managing any number of types of machines from a given control machine, so there's no point in evaluating local or remote results from the setup module to determine if role dependancies are satisfied.

mattwillsher:
Plus this module shouldn't care about wider package issues. I'll push up a new version in the next day or two. Thanks for your PR.

mattwillsher:
Sorry for the delay on getting this tagged & released. 0.3.0

PR #16, opened 2015-06-25
Make the role more container friendly

jitakirin:
Allow skipping service management entirely and ensure privilege separation directory exists.

This makes it more container friendly as service management isn't necessarily used there (especially in docker) and privilege separation directory is usually created by the service scripts on start.


jitakirin:
Tested on Ubuntu 14.04, CentOS 7 and Fedora 20 both in docker containers and in full VMs.  Behaviour on full VMs is backwards compatible with default settings.

mattwillsher:
Thanks!

PR #17, opened 2015-06-25
fix type in AcceptEnv

kostyrevaa:
default  AcceptEnv in RedHat includes LC_CTYPE but not LC_TYPE


mattwillsher:
Thank you  for the fix!

PR #21, opened 2015-06-26
Verify SSHd config early

jitakirin:
This uses the validate option to check the config file early, which
avoids putting bad settings in place at all, and also enables a
fail-fast behaviour (errors out when processing the template module).


mattwillsher:
Good call, thanks!

PR #30, opened 2015-11-05
Feature add: generate host keys with ssh-keygen

johanfleury:
This commit adds two variables that manage the generation or regeneration of SSH host keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,18 @@ If set to False, a reload of sshd wont happen on change. This can help with
 troubleshooting. You'll need to manually reload sshd if you want to apply the
 changed configuration. Defaults to the same value as ``sshd_manage_service``.
 
+* sshd_generate_host_keys
+
+If set to True, host keys will be generated with ``ssh-keygen -A``. Defaults to
+*False*.
+
+* sshd_regenerate_host_keys
+
+This options implies ``sshd_generate_host_keys``.
+
+If set to True, host keys will be deleted and and rengenerated with
+``ssh-keygen -A``. Defaults to *False*.
+
 * sshd
 
 A dict containing configuration.  e.g.
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -9,6 +9,10 @@ sshd_manage_service: "{{ false if ansible_virtualization_type == 'docker' else t
 sshd_allow_reload: "{{ sshd_manage_service }}"
 # If the below is false, don't manage /var/run/sshd directory
 sshd_manage_var_run: "{{ false if ansible_os_family == 'RedHat' and ansible_distribution_major_version == '7' else true }}"
+# Don't generate host keys
+sshd_generate_host_keys: false
+# Don't regenerate host keys
+sshd_regenerate_host_keys: false
 # Empty dicts to avoid errors
 sshd: {}
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -34,6 +34,14 @@
   tags:
     - sshd
 
+- name: Remove existing host keys
+  shell: rm -f /etc/ssh/ssh_host_*
+  when: sshd_regenerate_host_keys
+
+- name: Generate host keys
+  shell: ssh-keygen -A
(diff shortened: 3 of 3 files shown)


mattwillsher:
Hi Johan,

I like the idea of being able to regenerate the host keys. 

In my experience, at least for EL and Debian based distros, the sshd startup script generates the SSH host keys. 

It's my understanding that after keys are generated sshd will need to be reloaded/restarted to pick up the change.

Given these, I think the default action for regenerate should be to delete the keys via a call to a suitable module and call the reload/restart handler. The regeneration of the keys via ssh-keygen can be done iff a force generation boolean is set. 

Does this make sense and in line with your aims here?

Anthony25:
Hello @mattwillsher,
I worked with @johanfleury on this.

> @mattwillsher:
> Given these, I think the default action for regenerate should be to delete the keys via a call to a suitable module and call the reload/restart handler.

I did not understand what you mean by "a suitable module". You mean writing a module to replace the `shell` directive ?

mattwillsher:
Hi Anthony,

Thanks for getting back to this PR, I appreciate your contribution. I was considering incorporating the refinements myself, but it's your idea and it's good to see others shape the role.

Yes, I mean something other than shell - where there is a module that can perform a given action I like to use that rather than shell out. From memory I think this was a file delete? ( I'm on my phone so can't check right now) in which case the file module will do it, I think. 

It'd be also helpful to get some checks in the travis build configuration, but that definitely a stretch goal :)

If you need any assistance in getting this PR over the line please ask. Once it's ready I'll get it merged and a new version up on Galaxy.

Anthony25:
Ok, I see. A problem here is that file does not handle wildcards, so we cannot write:
```
file: dest="/etc/ssh/ssh_host_*" state=absent
```
That means we have to manually declare each hostkey if we still want to use `file`, which is (in my opinion) less permissive to future changes done in ssh, compared to `ssh-keygen -A` (the hostkey list will need to be up-to-date).
Using the workaround `shell: "rm /etc/ssh/ssh_host_*"` will not be a solution here… because we are still using shell! :p

@mattwillsher, what is your opinion about it?

mattwillsher:
Yes, I think find with register and then file would be succinct and suitable.

Another option is to look at the host keys var (sshd_defaults.HostKey) in role and use that. It's unlikely (though possible...) some one would override that value, as the intention is that that var holds the keys used by the given OS variant. To make it entirely generic I don't believe is possible - for example, ssh will no doubt introduce new parameters which will have to be encoded as defaults and also in the sshd template.

johanfleury:
Hi.

The find module is not available before ansible version 2.

I propose to use `command: find /etc/ssh -iname "ssh_host_*"` until ansible 2 is out and globally used.

If that's ok for you, i'll push my changes.

mattwillsher:
@johanfleury My preference is to use sshd_defaults.HostKey I think

tterranigma:
@mattwillsher why was this marked as `wontfix`?

jdnrg:
on amazonlinux2 validation fails due to missing hostkeys
```
TASK [willshersystems.sshd : Configuration] ************************************
fatal: [localhost]: FAILED! => {"changed": false, "checksum": "cb5869e74644dc696113badb660738991e077baa", "exit_status": 1, "msg": "failed to validate", "stderr": "Could not load host key: /etc/ssh/ssh_host_rsa_key\r\nCould not load host key: /etc/ssh/ssh_host_dsa_key\r\nCould not load host key: /etc/ssh/ssh_host_ecdsa_key\r\nCould not load host key: /etc/ssh/ssh_host_ed25519_key\r\nsshd: no hostkeys available -- exiting.\r\n", "stderr_lines": ["Could not load host key: /etc/ssh/ssh_host_rsa_key", "Could not load host key: /etc/ssh/ssh_host_dsa_key", "Could not load host key: /etc/ssh/ssh_host_ecdsa_key", "Could not load host key: /etc/ssh/ssh_host_ed25519_key", "sshd: no hostkeys available -- exiting."], "stdout": "", "stdout_lines": []}
```

PR #31, opened 2016-01-09
Fedora HostKey(s)

martinstefany:
Hi,

as mentioned in the commits, Fedora crew currently supports only F22 and F23, while F21 went EOSL by Dec 1 2015. Both F22 and F23 ship with openssh 7.1p1 which supports (and sets by default) 'rsa', 'ecdsa' and 'ed25519' HostKey options.

I'm using your ansible-sshd role (thank you very much!) on Fedora 23 and I had to do 'override' for this, so suggesting upstream change. Adjusting Fedora.yml seems more reasonable to me than to create another Fedora_23.yml, please, consider.

Thank you.
M.


mattwillsher:
Thanks @martinstefany . I'll release a new version soon.

Issue #32, opened 2016-01-11
"UnboundLocalError: local variable 'l_value' referenced before assignment"


stxyzth:
I am getting this error when running the default playbook example provided on your README:

```
TASK: [willshersystems.sshd | Configuration] ********************************** 
fatal: [NAP1] => {'msg': "UnboundLocalError: local variable 'l_value' referenced before assignment", 'failed': True}
fatal: [NAP1] => {'msg': "UnboundLocalError: local variable 'l_value' referenced before assignment", 'failed': True}

FATAL: all hosts have already failed -- aborting
```

mattwillsher:
Hi @stxyzth thanks for the report. Which version and Ansible and what OS and architecture?

stxyzth:
ansible 1.9.4
CentOS 6.7 final 64bit

mattwillsher:
Initial guest is that    `` sshd_skip_defaults: true`` is the problem and that the template expects something from there. I'd be interested to hear if setting that false helps.

I'm setting up my own test env. in any case. Not having the example working is a little embarrassing :)

mattwillsher:
I've tried this on fresh install of 6.7 x86_64 with Ansible 1.9.4 and don't get the error. I do get a problem with sandbox being set, so I've adjusted the README, and added a warning in #30 as the example will break SSH access to the server :)

stxyzth:
Any way I can output more information? I am still getting the issue despite doing:

```
    sshd_skip_defaults: false
    sshd_UsePrivilegeSeparation: no
```

mattwillsher:
Add -v to the command line, more v for more verbose up to -vvvv.

It may be worth posting an issue to the Ansible mailing list, or IRC channel, or try a clean install of Ansible. There was a bug in an earlier 1.9 version that looked a bit like this, but that was fixed in 1.9.3.

At this point, I'm not sure whether this is a bug in the role, in Ansible, or something peculiar to your set up.

luto:
I'm hitting the same error with [Ansible fb797a9](https://github.com/ansible/ansible/tree/fb797a9) (pre2.1.0) on CentOS 6.7, Python 2.6.6.

```
TASK [sshd : Configuration] ****************************************************                                                                                                                                                      
task path: roles/sshd/tasks/main.yml:37
<HOST> ESTABLISH SSH CONNECTION FOR USER: None
[...]

PR #35, opened 2016-01-24
Fix for CentOS 6 l_value issue

mattwillsher:
See #32


templates/sshd_config.j2
```diff
@@ -22,8 +22,12 @@
 {%     set value = sshd[key] %}
 {%   elif sshd_defaults[key] is defined and sshd_skip_defaults != true %}
 {%     set value = sshd_defaults[key] %}
+{%   else %}
+{%     set value = undefined %}
 {%   endif %}
+{%   if value is defined %}
```
luto: you don't need this, because it's already checked [inside `render_option()`](https://github.com/willshersystems/ansible-sshd/blob/lvalue_fix/templates/sshd_config.j2#L3).

meta/10_top.j2
```diff
@@ -22,8 +22,12 @@
 {%     set value = sshd[key] %}
 {%   elif sshd_defaults[key] is defined and sshd_skip_defaults != true %}
 {%     set value = sshd_defaults[key] %}
+{%   else %}
+{%     set value = undefined %}
 {%   endif %}
+{%   if value is defined %}
```
luto: https://github.com/willshersystems/ansible-sshd/pull/35/files#r50634826

luto:
works for me

mattwillsher:
Thanks for testing @luto and the initial PR . I'll use the approach you took in #34 as it just works.

luto:
Thank you for the speedy merge. Will you push a release, too? :)

mattwillsher:
Yes, all done 0.4.2 is up on galaxy now :)

Issue #38, opened 2016-03-01
Deprecation Warning


dphilpot:
Hi Willsher,

with the latest ansible version (2.0.1.0) we get a deprecation warning.

example:
`[DEPRECATION WARNING]: Using bare variables is deprecated. Update your playbooks so that the environment value uses the full variable syntax ('{{sshd_packages}}'). This feature will be removed in a future
 release. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.`

mattwillsher:
0.4.3 released with the fix from @haraldkoch

Thanks for the report.

PR #39, opened 2016-03-08
fix deprecation warning for sshd_packages (fixes issue #38)

haraldkoch:
This change fixes the ansible deprecation warning for bare word "sshd_packages" (issue #38).

(I tripped over it independently, but then found the issue already logged :)


mattwillsher:
Thanks!

PR #41, opened 2016-08-03
show xenial support on galaxy

snoopdouglas:
This role is 16.04 compatible but isn't showing up as such on Galaxy.

Have updated `meta/main.yml` so this is no longer the case.


mattwillsher:
Thanks!

snoopdouglas:
:)

PR #44, opened 2016-10-19
Support for OpenBSD

jasperla:
This was tested on OpenBSD 6.0 and `-current` with Ansible 2.1.2.0 and 2.2.0.0-rc1.


mattwillsher:
Thanks!

Issue #43, opened 2016-10-02
sshd_manage_var_run not necessary on Fedora


exploide:
I do not know why it is necessary for this role to explicitly manage the directory `/var/run/sshd`. But apart from that, there is an exclusion for RHEL7 but not for Fedora. I checked that a default installation of Fedora with working sshd does not have such a directory.

mattwillsher:
There was a reason for the management of /var/run/sshd, though I can't recall it through the mists of time and it may be that it was a bone headed call on my part.

I think I'll remove it entirely when I get time. For now, you can set sshd_manage_var_run to false as a role or inventory variable.

mattwillsher:
Now I remember. `/var/run/sshd` management is there for Docker. Much of the initial dev was in Docker and I moved to Vagrant VMs for additional testing. Under Docker, I seem to recall `/var/run` didn't exist. The [`file`](http://docs.ansible.com/ansible/file_module.html) module creates them recursively.

I'll review this again under #45 and sort it one way or another.

Issue #42, opened 2016-08-30
Circular symlink for testing


laneatomic:
The circular symlink in `tests/roles` causes copying the role to another server to break, especially with packer builds. Why not use an ansible.cfg file in `tests/` with the following contents:

```
[defaults]
roles_path = ../../
```

mattwillsher:
Good call - thanks for the report. I'll take a look when I get the time and get it fixed.

PR #45, opened 2016-11-01
test against multiple distros

chrismeyersfsu:
Use Docker in TravisCI to test multiple supported distros.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ script:
   - "ansible-playbook -i tests/inventory tests/test.yml --syntax-check"
 
   # Run the role for multiple distros
-  - "ansible-playbook -i tests/inventory tests/test.yml"
+  - "ansible-playbook -D -i tests/inventory tests/test.yml"
 
   # Run the role/playbook again, checking to make sure it's idempotent.
   - >
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,34 @@
 ---
-language: c
 sudo: required
+language: python
+python: "2.7"
+services:
+  - docker
 
 notifications:
     webhooks: https://galaxy.ansible.com/api/v1/notifications/
 
 before_install:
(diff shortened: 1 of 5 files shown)


mattwillsher:
Thanks for this, it's great to test against more operating systems. Interesting approach too.

I recall an issue with /var/run/sshd not existing inside Docker containers, hence the addition of /var/run/sshd management in this role. With the init system inside the containers, is this an issue? It looks like sshd might not be getting started during testing?

Looking at the output of the most recent run, sshd doesn't appear to be getting reloaded at the end despite changes to the sshd_config file. Is this an artefact of running inside Docker? The validation check looks to be passing though. 

I've added the -D option to the full run so, with a bit of luck, deltas for the config changes during testing can be seen.

chrismeyersfsu:
Thanks for the background info. A little background for you. The Docker containers that these tests against come with the init system shoved back in them and ssh running. Maybe I need to remove ssh.

I'll try running these tests locally and see how the role is interacting with the `sshd_manage_service` variable.

mattwillsher:
Good spot on sshd_managed_service. 

I recall having quite a bit of fuss with Docker at the time. `sshd_manage_service` and `sshd_manage_var_run` are a result. A few issues have been raised especially around `sshd_manage_var_run`, for example the currently open #43.

Do you have links to the container Dockerfiles? I had a look on Docker Hub but couldn't see the file or links to the RCS repository.

There are a couple of issues - #42 and #43 - that are related to Docker and testing that'd it'd be good to put to bed.

chrismeyersfsu:
Dockerfiles are here https://github.com/chrismeyersfsu/provision_docker/tree/master/files

mattwillsher:
Just to say I've not forgotten about this. Will get to work on it in the new year.

mattwillsher:
Two New Years past :( 

It's going to happen I'm afraid.

PR #49, opened 2017-02-11
clean Archlinux support to match the current package (openssh-7.4p1-2)

haraldkoch:
This is mostly a cosmetic change. It modifies the role to match the current OpenSSH version on Archlinux (7.4p1-2 as of this writing):

- automatically disable management of /var/run/sshd (not used on Archlinux)
- remove variables set in vars/Archlinux.yml that are not currently set in the distributed config
- remove sshd_service set in vars/Archlinux.yml because it is identical to defaults/main.yml


haraldkoch:
I hope you find this useful!

mattwillsher:
That's great, thanks!

mattwillsher:
That's great, thanks!

mattwillsher:
v0.4.8 is now on galaxy with this fix.

PR #50, opened 2017-03-16
Fix sshd service state

fuegoel:
I changed the state of the sshd service so if I rerun the playbook again and again, it will still work; otherwise, it will be broken at that step.


mattwillsher:
Thanks!

PR #51, opened 2017-03-17
Fix handler state

fuegoel:
Hi, me again. I've been facing this issue.
Symptom: 
- On the first run, the status of the play _Service enabled and running_ is **ok**, and the status of the handler _reload_sshd_ is **changed**. When I check on the target node, sshd is inactive (dead).
- On the second run, the status of the play is **changed**; sshd is active (running) now, but there's some error 

> error bind to port 22 on 0.0.0.0 failed address already in use 
- I tested manually on some random node and found out that if I _reload_ sshd when it's running, it'll always fail. But if I _restart_ it, it'll always succeed.
So this is my proposed fix. We're using CentOS here, and it works fine for us. I hope it's also ok for other distros.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -2,5 +2,5 @@
 - name: reload_sshd
   service:
     name: "{{ sshd_service }}"
-    state: reloaded
+    state: restarted
   when: sshd_allow_reload


mattwillsher:
Interesting...
Was sshd running before you ran the role? What changes are been made? Can you post your variables and playbook?

Reload is intended to only reload configuration for a running server and only calls  systemd reload . If that's failing on a running server there is something else at play here. Reload shouldn't start a new sshd.

I don't really want to change the reload handler to a restart because for most functions it should work as expected - asking sshd to reload its config.

fuegoel:
We're building VMs from the CentOS qcow image (the one for OpenStack), so yes, sshd is running from the very beginning.
This is our desired ssh config before the service gets reloaded:

>   PermitRootLogin: no
>   PasswordAuthentication: no
>   ChallengeResponseAuthentication: no
>   GSSAPIAuthentication: no
>   UsePAM: no
>   ClientAliveInterval: 30
>   ClientAliveCountMax: 99
>   AllowGroups: "{{ admingroup }}"
The playbook is a long one since it calls lots of roles. Basically, it builds a VM from the qcow image, deploys users, configures ssh, and installs and configures docker.
Can it make a difference because sshd is already running?

mattwillsher:
sshd running is the normal state and the state under which it was tested - after all, Ansible connects to the box using SSH.

```
error bind to port 22 on 0.0.0.0 failed address already in use
```

is something listening on the port. Is AllowGroups getting set correctly in the sshd_config file on first run? Does the config pass sshd -t? What do logs show (try ramping up the debug level)?

reload of sshd without Ansible should always work assuming `sshd -t` passes the config as being valid. If that's not the case, I'd say there is something in the CentOS image stopping it working as expected.

fuegoel:
`sshd -t` gives me `WARNING: 'UsePAM no' is not supported in RHEL and may cause several problems.` So it turns out that UsePAM being disabled caused the trouble. My bad, sorry.

mattwillsher:
Glad you found a fix.

It's also a bug on the roles part - it should run sshd -t before attempting a reload, though I did think that was being done. I'll raise an issue to check that.

PR #54, opened 2017-04-07
Fixed sshd_match blocks

troyfontaine:
This fixes the sshd_match blocks to use the contents of the matching block rather than re-using the original sshd_match block repeatedly.


mattwillsher:
Thanks! That's a pretty big bug there :)

PR #64, opened 2017-09-06
Makes handler use listen:

tterranigma:
Using the `listen:` option in handlers has the benefits of producing more user friendly output (task names) and allowing multiple handlers listen to the same event.


mattwillsher:
I didn't know that. Thanks.

PR #66, opened 2017-10-25
add default Ubuntu vars, prepend role_path for bug avoidance

moquist:
I'm trying to use Ubuntu 17.10 (which was not supported), and another role's vars/ kept being evaluated even after I added Ubuntu.yml. These changes fix both of those issues.

Thanks for considering these changes!

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -2,11 +2,11 @@
 - name: Set OS dependent variables
   include_vars: "{{ item }}"
   with_first_found:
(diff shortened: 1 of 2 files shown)


mattwillsher:
Thanks for the PR and sorry for the delay.

Not have a default for OS versions was (for better or worse) a design decision to ensure the OS always gets a valid SSH config.  Would calling the file Ubuntu_17.yml work for you?

I'm a bit perturbed about the other role's vars getting evaluated - that sounds like an Ansible bug. Which version of Ansible? What other role got evaluated?

PR #67, opened 2017-10-26
Fix Arch Linux var file

pieterlexis:
Hi,

My `ansible_os_family` var had the value `Arch Linux`. I symlinked the existing vars file accordingly.


mattwillsher:
Thanks for the PR!

PR #68, opened 2017-11-29
 adds absent option for deprecated value in Debian Stretch

slgevens:


mattwillsher:
Thanks for the PR, sorry for the delay.

I'm not quite sure of the use case here. Does there need to be a new defaults for stretch? Can you give an example of how this is used (and up date the readme if needed).

mattwillsher:
Closing due to lack of updates.

PR #70, opened 2018-02-06
Update default vars for Suse

dev-zero:
Based on SLE11/SP2 and a fresh installation of openSUSE Leap 42.3

### DIFF
--- a/vars/Suse.yml
+++ b/vars/Suse.yml
@@ -3,23 +3,13 @@ sshd_packages:
   - openssh
 sshd_sftp_server: /usr/lib/ssh/sftp-server
 sshd_defaults:
-  HostKey:
-    - /etc/ssh/ssh_host_rsa_key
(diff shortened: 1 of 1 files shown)


mattwillsher:
Thanks for the PR!

SUSE really disables PasswordAuthentication? How do users login after initial install?
Can the file be versioned (like for RedHat and Ubuntu) so older versions don't get changed?

mattwillsher:
Closing due to lack of update.

dev-zero:
Yes, Suse (and also other distributions which setup PAM by default) usually disable `PasswordAuthentication` and rely only on the `KeyboardInteractive`  since users could otherwise still login despite their account being disabled by PAM-specific means. See also https://superuser.com/q/894608 for an explanation of the difference.

Whether or not the file can be versioned is up to you: of course it can, but since your deployed configuration could allow users to bypass restrictions implemented via PAM I wouldn't recommend it.

PR #71, opened 2018-02-26
Add configuration append capability

jpiron:
To allow the use of other roles that update SSH configuration (like sftp
roles) while preserving idempotency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,11 @@ If set to False, a reload of sshd wont happen on change. This can help with
 troubleshooting. You'll need to manually reload sshd if you want to apply the
 changed configuration. Defaults to the same value as ``sshd_manage_service``.
 
+* sshd_append_configuration
+
+If set to True, configuration will be append to existing one instead of overriding it.
+This allows the use of other roles that update SSH configuration (e.g sftp roles, ...) while preserving idempotency.
+
 * sshd
 
 A dict containing configuration.  e.g.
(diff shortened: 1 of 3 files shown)


jpiron:
Any feedback on this ?

mattwillsher:
Hi,

Thanks for the PR. Been a bit slow with getting through these due to other commitments.

I have some questions, perhaps because I'm not quite grokking the code.

What is the reason for this change? Is it something that can't be done via the current templated system?
Can you give an example with code? 
Why does the default configuration get removed if the file is to be appended to?

jpiron:
Hi,
I am using your role along with the https://github.com/johanmeiring/ansible-sftp one which also updates the SSH configuration. With the template module, each time I run my playbooks the SSH role overrides the whole SSH configuration and then the SFTP role adds its configuration leading to Ansible reporting changes while nothing changed.
So the purpose of the modification is to allow the use of other roles that update SSH configuration while preserving idempotency.

As for the removal of the default configuration, this is because the blockinfile module will append the SSH configuration to the default one during the first run as it won't find the block markers.

mattwillsher:
The change makes me a little uneasy - there is a lot of additional complexity there. I think I'd prefer to see a entirely separate branch for append, touching the existing code minimally

Perhaps the other question is why what ansible-sftp does not be achieved by this sshd role? It should be feature complete for sshd.

mattwillsher:
Closing off as now has conflicts. Please feel free to resubmit if the mood takes you.

PR #76, opened 2018-04-16
Remove Deprecated options in default SSH config

TimJDFletcher:
OpenSSH developers have deprectated these options in OpenSSH 7.4 (which Debian 9 uses), this PR removes the options from the Debian9 config.

Stackexchange reference: https://unix.stackexchange.com/questions/337774/deprecated-options-when-restarting-openssh-in-stretch


mattwillsher:
Thanks for the PR!

Issue #77, opened 2018-04-19
Fix build


mattwillsher:
Build is failing saying ansible_virtualization_type is undefined.

D4N:
The build is green now, I guess this can be closed then?

Although not specific for this issue, but would it be possible to release a new version? The latest version of this role that one gets from ansible galaxy does not work with the most recent ansible version.

mattwillsher:
Fixed by #78

mattwillsher:
v0.6.0 has just been released. Thanks for making that possible.

D4N:
This is one of the fastest fixes I have encountered, thanks a lot!

PR #78, opened 2018-04-24
Fix for ansible_virtualization_type not being defined in Ansible > 2.5

D4N:
It looks like that `ansible_virtualization_type` is not defined by default in Ansible 2.5.1 when no virtualization is used. This resulted in this role failing due to an undefined variable.

This PR simply adds a jinja2 filter, so that `ansible_virtualization_type` defaults to `None` if not set.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -3,5 +3,5 @@
   service:
     name: "{{ sshd_service }}"
     state: reloaded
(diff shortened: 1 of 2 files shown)


D4N:
FYI: this should also fix the Travis build. The last one failed exactly due to the issue that this PR addresses.

mattwillsher:
Fixes #77 - Thanks very much!

PR #79, opened 2018-05-22
Amazon var name should be sshd_defaults

beyondbill:
Otherwise `sshd_skip_defaults: true` will not skip default settings on Amazon Linux.
This seems a typo since any other Linux distro's var file uses `sshd_defaults` rather than `sshd`.
If the fix makes sense, I hope to see a patch release soon since it currently blocks me in my work 🙏

### DIFF
--- a/vars/Amazon.yml
+++ b/vars/Amazon.yml
@@ -4,7 +4,7 @@ sshd_packages:
   - openssh
   - openssh-server
 sshd_sftp_server: /usr/libexec/openssh/sftp-server
-sshd:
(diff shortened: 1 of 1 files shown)


mattwillsher:
Thanks for the PR. I'll get a new release out shortly. Sorry it's taken 2 weeks.

mattwillsher:
v0.6.1 has been uploaded to Galaxy.

beyondbill:
@mattwillsher Thanks for the new release!

PR #83, opened 2018-07-03
Added TrustedUserCAKeys to options_match

maus-:
Pretty self explanatory diff, just adding TrustedUserCAKeys as a match option.


pillarsdotnet:
This is included in #87.

pillarsdotnet:
@mattwillsher  -- This should be closed, as it was included in #87

maus-:
included with #87

PR #84, opened 2018-07-27
expose sshd_config template backup option with sshd_backup variable

easonand:
I would like to have a backup of the config files that are overwritten when the template is copied to a machine with an existing config file.  I propose to add a new variable, sshd_backup, which defaults to false.  If you set sshd_backup to true, it will activate the template module's backup option.


mwillsher:
Thanks for the PR. I see no problem with this.

mattwillsher:
Thanks!

PR #85, opened 2018-08-09
Enforce CSV format for multi-value options where required.

pillarsdotnet:
Some sshd options require multiple-values to be expressed as a CSV list:
* `AuthenticationMethods`
* `Ciphers`
* `HostbasedAcceptedKeyTypes`
* `HostKeyAlgorithms`
* `KexAlgorithms`
* `MACs`
* `Match`
* `PubkeyAcceptedKeyTypes`

Although the existing code parses any `sshd_*` value which is expressed as a list, it fails to join list items with commas.  This PR ensures that multiple values associated with the above options are output as a one-line CSV list, whereas other (non-list) values are output one-per-line.


pillarsdotnet:
Apparently, travis-ci cannot contact ppa.launchpad.net, which is causing tests to fail because some of the code will not run in the version of ansible available in Ubuntu trusty repos.

This appears to be a periodic problem with travis-ci infrastructure. See for example [travis-ci#8834](https://github.com/travis-ci/travis-ci/issues/8834), [travis-ci#9112](https://github.com/travis-ci/travis-ci/issues/9112), and [launchpad#663396](https://answers.launchpad.net/launchpad/+question/663396).

pillarsdotnet:
Added a test for ansible version 2.0 or greater.

mwillsher:
Thanks for the PR and sorry for the delay in review - it's holiday time for me :)

The intent is that the value would be populated at comma separated - the module handles simple key/value pairs by design. This was so that new options could be added easily without understand what the options were as the role developer. I understand the desire to accept lists and parse them to comma separated lists, however. 

My initial though is that if a list is provided, it should be parsed to CSV, regardless of it being valid or not - the sshd config test will catch syntax value. 

The format of the options list and match causes me some pause - the options list was simple generated from the man page for the version of ssh at the time, with the expectation that adding new options would simply be a case of parsing the man page every so often. 

Give how significant the change is, the other, unrelated changes should be split out. 

The change the package installation systems needs separate justification.

mattwillsher:
Thanks for the PR and sorry for the delay in review - it's holiday time for me :)

The intent is that the value would be populated at comma separated - the module handles simple key/value pairs by design. This was so that new options could be added easily without understand what the options were as the role developer. I understand the desire to accept lists and parse them to comma separated lists, however.

My initial though is that if a list is provided, it should be parsed to CSV, regardless of it being valid or not - the sshd config test will catch syntax value.

The format of the options list and match causes me some pause - the options list was simple generated from the man page for the version of ssh at the time, with the expectation that adding new options would simply be a case of parsing the man page every so often.

Give how significant the change is, the other, unrelated changes should be split out.

The change the package installation systems needs separate justification.

pillarsdotnet:
@mattwillsher -- I have separated out the less-controversial changes into #87

I will look at re-working the option lists to avoid the second "delimited" column.  Maybe I can create a third options-list of configuration-keys which accept comma-delimited lists of values.

pillarsdotnet:
@mattwillsher -- Please note that the sshd config test does *not* catch the following error, for example:

```
MACs hmac-sha-512-etm@openssh.com
MACs hmac-sha2-256-etm@openssh.com
MACs umac-128-etm@openssh.com
MACs hmac-sha2-512
MACS hmac-sha2-256
MACs umac-128@openssh.com
```

When the above `/etc/ssh/sshd_config` configuration was generated, sshd applied the first `MACs` line and ignored subsequent `MACs` lines, which effectively prevented me from logging into the server via PuTTY.

pillarsdotnet:
The `HostKey` option, on the other hand, should *not* be specified as a comma-delimited list.  It should be specified as multiple single options, as:
```
HostKey /etc/ssh/ssh_host_rsa_key
HostKey /etc/ssh/ssh_host_dsa_key
HostKey /etc/ssh/ssh_host_ecdsa_key
HostKey /etc/ssh/ssh_host_ed25519_key
```

So we have some `sshd_config` options which must be specified as comma-delimited, and some which should not be so specified.  And we have at least one case where choosing incorrectly is not detected as an error by the sshd config test.

pillarsdotnet:
I just tested with an `/etc/ssh/sshd_config` file containing the following:
```
HostKey /etc/ssh/ssh_host_rsa_key,/etc/ssh/ssh_host_ecdsa_key,/etc/ssh/ssh_host_ed25519_key
```
Running `sshd -T` exited with success code 0.

But trying to connect resulted in the following logged error:
```
error: Could not load host key: /etc/ssh/ssh_host_rsa_key,/etc/ssh/ssh_host_ecdsa_key,/etc/ssh/ssh_host_ed25519_key
```

pillarsdotnet:
So now we have two cases:
* One where *failing* to join list items for an option that requires CSV format results in an unusable sshd configuration, with no warnings from Ansible.
* One ware *joining* list items for an option that disallows CSV format results in an unusalbe sshd configuration, with no warnings from Ansible.

In both cases, my PR above would have quietly done the correct thing.

mattwillsher:
This PR still contains the changes in #87

pillarsdotnet:
@mattwillsher -- Yup; let's get #87 merged first, then I'll rebase this one so we can discuss how to test.

mattwillsher:
#87 merged :)

mattwillsher:
For the tests I think the following config should be tested against:

- HostKey
- At least one key that takes a CSV value
- Other standard values

There is a case here for generally improved tests (actually make sure ssh is running and can be connected to), but that's a different matter.

pillarsdotnet:
Rebased.

Working on tests....

pillarsdotnet:
Gonna have to put this on hold for a few days; ran out of time. :-(

tterranigma:
@pillarsdotnet The error is a typo on [this line](https://github.com/willshersystems/ansible-sshd/pull/85/files#diff-3aef171dd3611e8720dbd8a65da064beR18). 

Nice approach. Some things I noticed.

The custom filter returns an empty string when the variable is not defined. This means that template will output all possible config options (as the are _defined_) with a value of `''`. Correct me if I'm wrong. 

But I don't understand what necessitates the use of a custom filter. There is the [`vars' lookup](https://docs.ansible.com/ansible/latest/plugins/lookup/vars.html) that can be used. 

Also, the template can be a bit faster by not calling the `render_option` macro for every possible option:

```jinja2
{% for option in sshd_options_body %}
{%   set optname = "sshd_"~option %}
{%   set optval = lookup('vars', optname, default='__undefined__' %}
{%   if optval == '__undefined__ %}
{{ body_option(option,optval) -}}
{% endif %}
{% endfor %}
```

One more thing, can you explain the `-` before some `}}`? Is this supposed to suppress newlines? I thought this only worked with `{%-`, `-%}`.

Issue #80, opened 2018-06-05
Support for multiple ssh daemons


tterranigma:
I have a need to configure and run 2 ssh daemons / services on different ports and generally different configuration. AFAICT this is not possible currently with this role. Would there be interest to add this functionality? I could work on a first implementation.

mattwillsher:
I'm unclear on the use case for this. sshd supports Match blocks to adjust configuration based on User, Group, Host, and Address and individual users can have restrictions placed in their authorized_keys files.

I'm happy to consider adding the functionality if you can clearly justify the case.

easonand:
Why not just include a separate play in your playbook with sshd_config_file, sshd_binary, sshd_service, etc. set?

---
- hosts: all
  vars:
    sshd_skip_defaults: true
    sshd:
      Compression: true
      ListenAddress:
        - "0.0.0.0"
        - "::"
      GSSAPIAuthentication: no
      Match:
        - Condition: "Group user"
          GSSAPIAuthentication: yes
    sshd_UsePrivilegeSeparation: no
    sshd_match:
        - Condition: "Group xusers"
          X11Forwarding: yes
  roles:
    - role: willshersystems.sshd

# it is fine to have more than one play in a playbook
- hosts: somehost
  vars:
    sshd_binary:
        /usr/local/sbin/sshd
    sshd_config_file:
       /usr/local/etc/ssh/sshd_config
    sshd_service:
       special-ssh
    sshd:
         Port: 
            - 9999
            - 2222
  roles:
    - role: willshersystems.sshd

tterranigma:
@mattwillsher I cannot use Match rules as I need to have different ports for the two sshd instances.

@easonand Thank you for your suggestions. They got me thinking and I worked on #89 . The problem I encountered with your approach was that the systemd service cannot be overridden by the role user, it is hard set once depending on the OS.

Let me know what you think of #89.

mattwillsher:
Fixed by #89 and in v0.7.0

PR #89, opened 2018-08-25
Adds ability to install a systemd service

tterranigma:
This PR is the result of me trying to solve #80 . The issue with the existing role is that if you want to install a second sshd service, it is not possible due to the variable `sshd_service` being "hardcoded". What happens is that when the OS-specific vars/ file is included, any custom value for that variable is overwritten. Also, it is not possible to use Match rules for everything as the directives supported there are limited (for instance the `Port` directive is not valid there and this is what I mainly wanted).

In this PR, I tried to solve this issue by adding an 'override' logic. If the user sets any of the vars/ variables they do not get overwritten by those under vars/. 

Also, this PR adds some templates for the sshd systemd service. I took them from Ubuntu 18.04 and simplified them. 

With this PR, I am now able to control 2 different sshd instances on my server, running on different ports.

I had only edited `Debian.yml` and `Ubuntu_18.yml`. If you agree with this PR, I will go the extra mile and edit the rest under vars/.


mattwillsher:
I'm generally in favour of this change as it makes the role more flexible.

It obviously needs to pass the build tests.

tterranigma:
I updated the rest of the vars/ files.

mattwillsher:
Thanks for that - can you also update the README to reflect this new functionality?

mattwillsher:
That's great, thanks. I'll get a new release out soon.

Jakuje:
@nkakouros Hi, can you clarify what was the reason you added the set_facts part (ideally with some example playbook)? This makes this role pretty much unusable if invoked multiple times from a single playbook as the facts are global and their scope goes out of the first role invocation into the next one, making it impossible to use the role to create different included drop-in files or configuring more sshd services.

From my understanding, Ansible has (at least today) quite a nice variable precedence table and this role is following that so I do not see a reason why it would be a problem to override some of the system defaults.

I submitted a PR to remove these because of the reasons given above, but before doing so I would be interested why it was added in the first place. The commit message is not very descriptive.

https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

PR #91, opened 2018-09-08
Adds on/off toggle

tterranigma:
This is something I do to all my custom roles. It allow for any easy on off switch and helps in testing playbooks and in creating more dynamic playbooks.


mattwillsher:
I don't see a conditional clause in the commit that would make this work

tterranigma:
Uhhhmmmmm, right.. :blush:

PR #92, opened 2018-09-08
Updates syntax to Ansible 2.7 era

nkakouros:
As ansible 2.7 is not out yet, if you are interested in this PR, maybe it should be accepted [after 2.7 is out](https://docs.ansible.com/ansible/devel/roadmap/ROADMAP_2_7.html).

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,7 +1,10 @@
 ---
+
 - name: Reload the SSH service
   service:
     name: "{{ sshd_service }}"
     state: reloaded
-  when: "sshd_allow_reload and ansible_virtualization_type|default(None) != 'docker'"
+  when:
+    - sshd_allow_reload
+    - ansible_virtualization_type|default(None) != 'docker'
   listen: reload_sshd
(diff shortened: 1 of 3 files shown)


tterranigma:
The failure is valid, I have [asked](https://github.com/ansible/ansible/issues/18430) on the ansible project on how to combine the new `loop` keyword with `first_found`.

mattwillsher:
The role should work on as many versions as reasonably possible, although the test is done against the current version of the ansible ppa.

At least, a few months should pass before mandating 2.7 as a minimal version.

tterranigma:
In that case, this should be put on hold until the `with_` is removed (I couldn't find a target release when this will happen). Should we close this now and retry then?

mattwillsher:
It can be left open, it's a good reminder.

tterranigma:
Now that 2.8 is out, I think it would be a good idea to merge this. Since the deprecation cycle in ansible is 4 minor versions long, the oldest "officially" supported version is 2.5 now. The changes in this PR, I think should be compatible with at as old a version as 2.5.

mattwillsher:
Merged, thanks. Will release with v0.8.0 shortly.

PR #93, opened 2018-09-10
Fixes bad option in systemd service file

tterranigma:
This produced a non-fatal error when the service was started.


mattwillsher:
Thanks. I'll do a release shortly.

PR #95, opened 2018-09-11
Make role work with chroot connections on EL 7.

jabl:
For chroot connection (e.g. when building images instead of working on
live hosts) one cannot restart services etc.

Also due to Ansible bug 21026 one must run systemctl instead of using
the service module, limiting support to EL 7 for the time being.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -41,7 +41,18 @@
     name: "{{ sshd_service }}"
     enabled: true
     state: started
-  when: "sshd_manage_service and ansible_virtualization_type|default(None) != 'docker'"
+  when:
+    - sshd_manage_service
+    - ansible_virtualization_type|default(None) != 'docker'
+    - ansible_connection != 'chroot'
+
+# Due to ansible bug 21026, cannot use service module on RHEL 7
+- name: Enable service in chroot
+  command: systemctl enable {{ sshd_service }}
+  when:
+    - ansible_connection == 'chroot'
+    - ansible_os_family == 'RedHat'
+    - ansible_distribution_major_version|int >= 7
 
 - name: Register that this role has run
   set_fact: sshd_has_run=true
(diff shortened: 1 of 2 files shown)


tasks/install.yml
```diff
@@ -41,7 +41,18 @@
...
     state: started
-  when: "sshd_manage_service and ansible_virtualization_type|default(None) != 'docker'"
+  when:
+    - sshd_manage_service
+    - ansible_virtualization_type|default(None) != 'docker'
+    - ansible_connection != 'chroot'
+
+# Due to ansible bug 21026, cannot use service module on RHEL 7
+- name: Enable service in chroot
+  command: systemctl enable {{ sshd_service }}
```
mattwillsher: Can the systemd module to used here? 

Could sensible_service_mgr fact to be used rather than the OS family and major version?

tasks/install.yml
```diff
@@ -41,7 +41,18 @@
...
+  when:
+    - ansible_connection == 'chroot'
+    - ansible_os_family == 'RedHat'
```
mattwillsher: Could sensible_service_mgr fact to be used rather than the OS family and major version?

tasks/install.yml
jabl: No, it also fails, with the (slightly non-sensible) error message "Failed to get D-Bus connection: Operation not permitted"

tasks/install.yml
jabl: Presumably you mean the ansible_service_mgr fact?  This was what I actually was searching for but failed to find, so I resorted to using the OS/version instead.

Unfortunately, I just tried it, and in a  chroot image it for some reason claims that "ansible_service_mgr == 'service'" rather than 'systemd' as it properly does on a live host.

mattwillsher:
Merged, thanks.

Jakuje:
Hi. I just stumbled upon this part of the source code and I wanted to check if it is still relevant as the referenced issue is marked as fixed:
https://github.com/ansible/ansible/issues/21026

PR #97, opened 2018-09-20
Add vars for openSUSE Leap 15 and CentOS 8

dev-zero:
Based on SLE11/SP2 and a fresh installation of openSUSE Leap 42.3

... rebased version of PR #70, including requested comments:

Suse (and also other distributions which setup PAM by default) usually disable PasswordAuthentication and rely only on the KeyboardInteractive method since users could otherwise still login despite their account being disabled by PAM-specific means. See also https://superuser.com/q/894608 for an explanation of the difference between Password and KeyboardInteractive authentication.

Whether or not the file can be versioned is up to you: of course it can (for example roll out the new version for `ansible_lsb.major_release|int >= 15 and <42`, just have to make sure that the lsb package is installed), but since your current configuration could allow users to bypass restrictions implemented via PAM I would not recommend it and instead put it in the release notes for the next version.

### DIFF
--- a/vars/RedHat_8.yml
+++ b/vars/RedHat_8.yml
@@ -1,0 +1,28 @@
+---
+__sshd_packages:
+  - openssh
+  - openssh-server
+__sshd_sftp_server: /usr/libexec/openssh/sftp-server
+__sshd_defaults:
+  HostKey:
+    - /etc/ssh/ssh_host_rsa_key
+    - /etc/ssh/ssh_host_ecdsa_key
+    - /etc/ssh/ssh_host_ed25519_key
+  SyslogFacility: AUTHPRIV
+  AuthorizedKeysFile: .ssh/authorized_keys
+  PasswordAuthentication: yes
+  ChallengeResponseAuthentication: no
+  GSSAPIAuthentication: yes
+  GSSAPICleanupCredentials: no
+# Note that UsePAM: no is not supported under RHEL/CentOS. See
+# https://github.com/willshersystems/ansible-sshd/pull/51#issuecomment-287333218
+  UsePAM: yes
+  X11Forwarding: yes
+  PrintMotd: no
+  AcceptEnv:
+    - LANG LC_CTYPE LC_NUMERIC LC_TIME LC_COLLATE LC_MONETARY LC_MESSAGES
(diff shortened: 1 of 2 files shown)


mattwillsher:
Suse doesn't set any HostKey values?

Can she ansible_distribution_major_version be used to makes sure these values only apply to the correct version of Suse?

dev-zero:
@mattwillsher I will soon install another one and check again, but the last ones I checked (SLES 12 SP3, Leap 15.0, Leap 42.3, Leap 13.1) did only set the variables given in this PR and otherwise rely on the compiled-in defaults of sshd. Do you know where those defaults were coming from?

Sure, you could use `ansible_distribution_major_version` to contain the new settings, although I'm not sure to what (`>= 15' would be the obvious choice, but that we leave SLES 12 out).

dev-zero:
Ok, I've verified again by looking at the source:

* https://build.opensuse.org/package/view_file/network/openssh/openssh.spec?expand=1 ... installs the original OpenSSH `sshd_config`
* OpenSSH does not set `HostKey`: https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/sshd_config?rev=1.103&content-type=text/x-cvsweb-markup
* openSUSE only changes `PermitRootLogin yes`, `UsePAM yes`, `X11Forwarding yes` and forward the following vars:

```
# This enables accepting locale enviroment variables LC_* LANG, see sshd_config(5).
AcceptEnv LANG LC_CTYPE LC_NUMERIC LC_TIME LC_COLLATE LC_MONETARY LC_MESSAGES
AcceptEnv LC_PAPER LC_NAME LC_ADDRESS LC_TELEPHONE LC_MEASUREMENT
AcceptEnv LC_IDENTIFICATION LC_ALL
```
... as per the patches in https://build.opensuse.org/package/show/network/openssh

By now I would probably use the following restriction to apply them:
`15 <= ansible_distribution_major_version < 42`

dev-zero:
in contrast to #116 I found `GSSAPICleanupCredentials no` on my freshly installed CentOS 8 machines.

mattwillsher:
Sorry for the delay. Merged. I'll get a new released out soon. Thanks for the update.

PR #104, opened 2019-02-25
Fix variable loading.

pillarsdotnet:
I discovered that [tasks/variables.yml](https://github.com/willshersystems/ansible-sshd/blob/master/tasks/variables.yml#L3) was searching in the `files` directory rather than the `vars` directory. (!!!)

This patch ensures that variable files are read from the `vars` directory of the current role or playbook.

It also adds compatibility for odd-numbered Ubuntu major releases.

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -1,7 +1,7 @@
 ---
 galaxy_info:
   author: Matt Willsher
-  description: OpenSSH SSH deamon configuration
+  description: OpenSSH SSH daemon configuration
   company: Willsher Systems
   license: LGPLv3
   min_ansible_version: 1.8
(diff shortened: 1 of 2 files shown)


pillarsdotnet:
See https://github.com/ansible/ansible/issues/23392 for context.

mattwillsher:
I'm a bit perplexed by this - is that to say it _never_ applies the correct variables based on OS? I think it must have working on some older release of Ansible, but changed?

tterranigma:
The issue is that ansible first looks into the tasks/ dir and then in the vars/ dir. I think the PR is good to have as it will also help in moving to the `loop` keyword.

Issue #105, opened 2019-03-22
Setting sshd_backup has no effect


BenGig:
As far as I can see, the functionality for _sshd_backup_ is not implemented

mattwillsher:
Fixed by MR #106

PR #106, opened 2019-03-22
Backup of sshd_config dependent on variable

BenGig:
Propagates setting sshd_backup to template installation task (issue #105)


mattwillsher:
Thank you

PR #107, opened 2019-04-29
Remove 'UsePrivilegeSeparation' from Fedora defaults

kpfleming:
This option has been deprecated in OpenSSH for nearly two years, was the default for five years before that, and is not part of the current Fedora default configuration. It should not be included in sshd_config on Fedora systems.


mattwillsher:
Thanks!

PR #112, opened 2019-07-10
AIX support (including new AIX handler)

d-little:
#### Description

Added a default var file (`AIX.yml`) for AIX defaults.  Updated the Readme to add additional limitations of AIX for the role.  

Added a new handler to restart sshd on AIX; `service` module does not support 'reload', and sshd requires some special handling to restart.

#### Type of change

- [X] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

#### How Has This Been Tested?

Tested on AIX 7.1 and 7.2 - I havent tested across the board functionality, but there shouldnt be an change in the role's behaviour on other environments.


d-little:
FWIW: Yep, on a fresh install, right out of the box, the only thing that AIX changes is sftp subsystem:
```shell
$ egrep -v '(^$|^#)' /etc/ssh/sshd_config
Subsystem       sftp    /usr/sbin/sftp-server
```

mattwillsher:
Thanks @d-little. Will release v0.8.0 with this change in shortly.

PR #114, opened 2019-07-18
Support for Debian Buster

yokozuna:
Add support for Debian Buster Debian_10.yml
Remove deprecated sshd option UsePrivilegeSeparation from Debian_10.yml


mattwillsher:
Thanks for your contribution. I merged #117 as that seemed more complete.

PR #115, opened 2019-09-30
OpenBSD and ansible_distribution_major_version

thalunil:
* OpenBSD no longer provides ansible fact "ansible_distribution_major_version" therefore set it to "ansible_distribution_version".

* ansible fact ansible_distribution_lts_version only makes sense on Ubuntu so set it only on Ubuntu -> otherwise "ansible_distribution_version"

### DIFF
--- a/tasks/variables.yml
+++ b/tasks/variables.yml
@@ -16,15 +16,12 @@
         if ansible_distribution == "Ubuntu"
         else ansible_distribution_version
       }}
-    ansible_distribution_major_version: >-
-      {{
-        ansible_distribution_version if ansible_os_family == "OpenBSD"
-      }}
     params:
       files:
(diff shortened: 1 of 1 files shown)


thalunil:
Please consider this pull request, this Ansible role is currently dysfunctional on OpenBSD.

Error:
TASK [sshd : Set OS dependent variables] ******************************************************************
fatal: [...]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ansible_distribution_major_version' is undefined\n\nThe error appears to be in '.../roles/sshd/tasks/variables.yml': line 3, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set OS dependent variables\n  ^ here\n"}

Thank you

mattwillsher:
Thanks. Will be in the next release (0.10.0)

PR #116, opened 2019-10-09
Adding support for RedHat8/CentOS8.

bengoa:
Adding support for RedHat8/CentOS8.

Dropping deprecated option `UsePrivilegeSeparation`.

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -22,6 +22,7 @@ galaxy_info:
     versions:
     - 6
     - 7
+    - 8
   - name: Fedora
     versions:
     - 22
--- a/vars/RedHat_8.yml
+++ b/vars/RedHat_8.yml
@@ -1,0 +1,27 @@
+---
+__sshd_packages:
+  - openssh
+  - openssh-server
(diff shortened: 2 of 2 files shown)


dev-zero:
on my fresh CentOS8 (minimal) installation, I have: `GSSAPICleanupCredentials no`

bengoa:
Hello,

On a fresh CentOS7 (minimal) installation you'll have `GSSAPICleanupCredentials no` as well, but this role is changing it to `yes`.

I tried to keep the same behaviour, dropping only the deprecated option, but I can change it and update here if you prefer.

dev-zero:
unless there's a good reason to diverge from the distro defaults (like the `PermitRootLogin`) I usually prefer to keep them. Since here you're adding new support for CentOS/RedHat 8, this might be a good point to bring the ansible-sshd configuration in line with the distro defaults.

bengoa:
All right, I agree with you. I'll push an update soon.

mattwillsher:
I took #97 over this. Thanks for you contribution.

PR #120, opened 2020-01-19
Add Gentoo support (with secure sshd defaults)

oddlama:
This PR adds support for all Gentoo systems. Also includes secure defaults for the sshd configuration for gentoo.


mattwillsher:
Thank you.

PR #121, opened 2020-02-14
End_host for unsupported hosts

Spreadcat:
This change will allow the play to continue without error if unsupported hosts are in the lists of targed host.
The play will continue with the supported hosts end the play for the ones which are not supported.

We had an issue where this role is used in a workflow against newly deployed nodes and obviously windows hosts will fail on this. This change allowed the workflow to continue.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -1,8 +1,8 @@
 ---
 
 - name: OS is supported
-  assert:
-    that: __sshd_os_supported|bool
+  meta: end_host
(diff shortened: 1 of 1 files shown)


mattwillsher:
Thanks!

ahmadalli:
this breaks playbooks running with ansible 2.7 :(

mattwillsher:
I should have checked the wider impact of this change.  I'll revert it for now.

PR #124, opened 2020-03-29
exit_host on ansible >= 2.8

ahmadalli:
I've added github action to lint the playbook but unfortunately it doesn't support testing with ansible 2.7 for now


ahmadalli:
I've added tests for ansible 2.7 to 2.10

mattwillsher:
Thanks. Sorry for the delay. 
The test failed for 2.7, though going forward I was planning on only support the currenly maintained Ansible version (2.8 currently).

PR #128, opened 2020-09-18
Remove test for Ansible 2.7

mattwillsher:
Ansible 2.7 is no longer maintained and ansible-lint requires >=2.8


mattwillsher:
Dupe of #129

Issue #140, opened 2020-09-25
CentOS 8 / EL 8 problem


lentferj:
sshd on CentOS8 seems to be not configured through /etc/ssh/sshd_config, but loads /etc/crypto-policies/back-ends/opensshserver.config via the systemd sshd.service configuration. So while you can use this role to manage /etc/ssh/sshd_config, it won't take effect.

lentferj:
The "problem" (if one like to call it that) seem to be "system-wide cryptographic policies" in EL8, which take precedence over some settings in sshd_config (ciphers, mac, etc). If you want to return the control back to sshd_config you need to uncomment the line with the CRYPTO_POLICY= variable in the /etc/sysconfig/sshd file according to the section OpenSSH on the lower half of this page
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening

Having an option to set this via the role would be nice

Jakuje:
The sshd is still configured through `sshd_config` in RHEL8, but only crypto-related configuration is overridden on the command-line by crypto policy to allow setting this system-wide.

I am not sure if this is something that should come here as if you are targeting some particular cryptographic configuration or compliance, you should update the crypto policies directly [1]. But I understand it can be handy and quite common request. What do you think @mattwillsher ?

[1] https://www.redhat.com/en/blog/how-customize-crypto-policies-rhel-82

lentferj:
I tried changing just the sshd settings with crypto-policies. If I have not overseen anything, if you want to change the settings for just one service, you have to set up your changes in /etc/crypto-policies/local.d, e.g. which in the case of sshd will then be appended to what is defined in /etc/crypto-policies/back-ends/opensshserver.config.

So, to change the crypto setup for opensshserver, I would add something like:

\# cat /etc/crypto-policies/local.d/opensshserver-local_conf.config 
[...]

Issue #141, opened 2020-10-03
Add support for /etc/sysconfig/sshd options for EL8


mattwillsher:
CRYPTO_POLICY in /etc/sysconfig/sshd is requested in #140 

Add support for anything else in there too.

PR #142, opened 2020-10-06
Support /etc/sysconfig/sshd to override crypto policies and handle more advanced use cases

Jakuje:
This should solve #141. Test playbook is written too, but it does not make much sense to run in on Ubuntu as it should not do anything there.

On the other hand, the file is just list of environment variables  so it should not hurt if unsupported values are written in there. I would be glad for review/comments/suggestions/testing.


richm:
lgtm

mattwillsher:
Closes #140, #141

PR #143, opened 2020-10-08
Implement more natural match blocks and test them

Jakuje:
One of the things that I noticed cumbersome in this role was the definition of Match blocks, which was supported in different ways, but limited to 11 blocks or so with `sshd['Match']`, `sshd_match` and `sshd_match_[1-9]`.

This implements alternative way describing match blocks as a list in `sshd['Match']` or `sshd_match`, keeping the old semantics unchanged for backward compatibility.


mattwillsher:
Thanks @Jakuje.

Issue #111, opened 2019-06-18
Configuration step fails for CentOS7 if no ssh packages were installed


ghost:
When using the role to install sshd in a centos7 Docker container the role installs the packages, but the sshd is never started before the "Configuration" step.

Due to this there are no hostkeys, which seem to be generated when sshd is started for the first time and there are no keys yet:
```TASK [willshersystems.sshd : Configuration] ************************************
fatal: [localhost]: FAILED! => {"changed": false, "checksum": "3eb484dee3405d49871d72d546e490cd0a6bf897", "exit_status": 1, "msg": "failed to validate", "stderr": "Could not load host key: /etc/ssh/ssh_host_rsa_key\r\nCould not load host key: /etc/ssh/ssh_host_ed25519_key\r\nsshd: no hostkeys available -- exiting.\r\n", "stderr_lines": ["Could not load host key: /etc/ssh/ssh_host_rsa_key", "Could not load host key: /etc/ssh/ssh_host_ed25519_key", "sshd: no hostkeys available -- exiting."], "stdout": "", "stdout_lines": []}
```

I thought about just fixing this, but this seems more complicated, as one would have to actually dig through the sshd_config.hostkey variable to find the expected name for the hostkeys and actually figure out the type for the hostkey and then generate them, if they do not exist.

Another idea would be to unconditionally start sshd before the configuration step, to ensure the existence of the hostkeys, but I do not know if this may create other problems.

This is a rather special corner case, as ansible works over SSH and so this will only ever be a problem, when setting up a container or any situation where the playbook does not rely on sshd actually running (ansible_connection=local e.g.).

If simply unconditionally starting sshd before the configuration step would be an option (or just doing this under certain conditions) I could come up with a quick PR.

Any ideas or suggestions?

pdreker:
This issue was opened by me - I had to shuffle GitHub accounts around and the original one went the way of the Dodo...

Another option would be to call "ssh-keygen -A" which will automatically generate all missing SSH host keys, which the installed version supports.

mattwillsher:
This was an issue from the early days of the role. I seem to recall I did try to generate the keys, but given the issues, as you state, I decided not to. 

Perhaps code could be added to generate keys based on data passed into the role? 
Alternatively, maybe the check could be optional? 

For my learning, what's the use case for installing ssh into a docker container?

pdreker:
What is the use case...? Good question ;-)
[...]

PR #146, opened 2020-11-06
Run tests with Github Actions and fix things on the way

Jakuje:
This is a bigger changeset which I put together while trying to run all the test with Github Actions on Fedora and CentOS 7 and 8. There were couple of issues that I try to fix in this, which might be more intrusive, but should not change the external API as far as I see. The summary of changes:

 * Getting rid of the `set_facts`, which was introduced in #89 and which is making the role unusable repetitively in a single play.
 * Implement check & generation for hostkeys, which should solve #111 
 * Many tests improvements
 * Improve/cleanup documentation
 * Remove some unneeded default options
 * Improved support for creating configuration snippets (for `Include` functionality default in Fedora)
 * Testsuite passing in Fedora, CentOS 7 and CentOS 8

It should be pretty easy to add Ubuntu and Debian, but their sshd service temporary directory is created only during start of the service (which is not happening in containers) so it will require some workaround even to the test mode (sshd -t).


mattwillsher:
This looks ok to me. I'll merge in a few days to see if there is response to your update to #89.  Once merged I'll tag for v0.12.0

At some point v1.0 should be considered - do we need a roadmap?

Jakuje:
Thank you. I am also waiting for internal review, so no need to hurry. I would be happy to accommodate existing use case of the user from #89, but I just do not see it now.

At this moment, I believe most of the things that we needed should be in and working, but certainly some issues or needs will show up in the meantime so I would certainly wait for v1.0 to get more testing from users. We have some internal uses too where I would like to gather feedback too.

README.md
```diff
@@ -6,12 +6,12 @@ OpenSSH Server
 This role configures the OpenSSH daemon. It:
 
 * By default configures the SSH daemon with the normal OS defaults.
-* Works across a variety of UN*X like distributions
+* Works across a variety of Unix-like distributions
```
richm, line 9: https://en.wikipedia.org/wiki/Trademark_Unix
I would say `*nix` instead of `UN*X` or `Unix-like`

defaults/main.yml
```diff
@@ -10,16 +10,16 @@ sshd_skip_defaults: false
 # daemon at all
 sshd_manage_service: true
 
+# If the below is false, don't reload the ssh daemon on change
+sshd_allow_reload: true
```
richm, line 14: What does this mean?  Does this mean that sshd will use something like inotify to watch for changes in its configuration files, and automatically restart or reload the configuration?

tasks/install.yml
```diff
@@ -32,6 +21,88 @@
...
+        validate: >-
+          {% if sshd_test_hostkey is defined and sshd_test_hostkey.path is defined %}
+            {{ sshd_binary }} -t -f %s -h {{ sshd_test_hostkey.path }}/rsa_key
+          {% else %}
+            {{ sshd_binary }} -t -f %s
+          {% endif %}
+        backup: "{{ sshd_backup }}"
+      notify: reload_sshd
+
+    - name: Remove temporary host keys
```
richm, line 99: this should be in an `always` block - https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html#handling-errors-with-blocks

tests/tests_alternative_file.yml
```diff
@@ -12,19 +12,78 @@
...
       - meta: flush_handlers
 
       - name: Print current configuration file
         command: cat /etc/ssh/sshd_config_custom
         register: config
 
-      - name: Check the options are in configuration file
+      - name: Print second configuration file
+        command: cat /etc/ssh/sshd_config_custom_second
+        register: config2
```
richm, line 55: FYI, for future code - this is ok for now - I see that the existing code uses `cat /remote/file` but the "ansible" way to do this is to use the `slurp` module https://docs.ansible.com/ansible/2.9/modules/slurp_module.html#slurp-module

vars/Fedora_31.yml
```diff
@@ -1,8 +1,8 @@
 ---
-__sshd_packages:
```
richm, line 2: Fedora 31 will soon be eol - do you still need a separate fedora_31 vars file?

Jakuje: There is a new support for sshd_config include in Fedora 32 so I wanted to keep this difference before it will be actually EOL. Then I think we can remove it altogether.

tasks/install.yml
richm: hmm - what version of ansible are you using?

defaults/main.yml
richm: no, just curious

README.md
richm: ok - how about ```UN*X``` - will that make the syntax highlighter work?

tasks/install.yml
Jakuje: This was running on Fedora 32 (ansible 2.9.14) with targets on CentOS 7 and CentOS 8 images using standard test inventory:
```
TEST_DEBUG=1 TEST_SUBJECTS=CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2 ansible-playbook -v -i /usr/share/ansible/inventory/standard-inventory-qcow2 tests_*.yml
```
Does it sound like something that should not happen?

There is a bit more logs:
```
TASK [ansible-sshd : Create a temporary hostkey for syntax verification if needed] **********************************
skipping: [CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2] => {"changed": false, "skip_reason": "Conditional result was False"}

TASK [ansible-sshd : Generate temporary hostkey] ********************************************************************
skipping: [CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2] => {"changed": false, "skip_reason": "Conditional result was False"}

TASK [ansible-sshd : Create the configuration file] *****************************************************************
fatal: [CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2]: FAILED! => {"changed": false, "checksum": "16d1d23e6df9619f96a5690ef7150237a627f79f", "exit_status": 1, "msg": "failed to validate", "stderr": "Unable to load host key: /tmp/missing_ssh_host_rsa_key\r\nsshd: no hostkeys available -- exiting.\r\n", "stderr_lines": ["Unable to load host key: /tmp/missing_ssh_host_rsa_key", "sshd: no hostkeys available -- exiting."], "stdout": "", "stdout_lines": []}

TASK [ansible-sshd : Remove temporary host keys] ********************************************************************
skipping: [CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2] => {"changed": false, "skip_reason": "Conditional result was False"}

TASK [Check that we failed in the role] *****************************************************************************
fatal: [CentOS-8-GenericCloud-8.2.2004-20200611.2.x86_64.qcow2]: FAILED! => {"msg": "The conditional check 'ansible_failed_result.msg != 'UNREACH'' failed. The error was: error while evaluating conditional (ansible_failed_result.msg != 'UNREACH'): 'ansible_failed_result' is undefined"}
```
Let me know you will need some different logs or minimal reproducer or if I do something wrong.

README.md
Jakuje: Enclosing with \` does not confuse vim so I will change it to it. Thanks for suggestion.

tasks/install.yml
richm: hmm - I think it may be because you have a nested block.  tasks/install.yml is doing the above tasks inside a block - I think the `ansible_failed_result` is scoped to that block, and perhaps is not set in the enclosing block.  I would think the enclosing block would also have that variable, and that it should be set since there was a failure from the block, but perhaps this is a bug in Ansible?  I guess the situation is analogous where you have nested exceptions and you want to capture the exception at the lower level and raise it again for the higher level?  e.g.
```
try:
  something()
  try:
    some_command_that_errors()
  except Exception as exc:
    do_something()
    raise exc
except Exception as exc:
  do_something()
```
the problem is that I don't know if there is something analogous to `raise` in Ansible which will propagate an error from a nested `block` to a higher level.

richm: https://richm.github.io/how-to-catch-and-reraise-errors-in-ansible

mattwillsher:
I'm going to convert this to a draft for now to avoid accidental merging.

tasks/install.yml
Jakuje: Thank you for pointing to the right direction. This is the exact use case we need. I just pushed updated PR.

Jakuje:
Updated from the draft as it looks like all the comments were resolved and CI is passing.

PR #147, opened 2020-11-24
Improve test coverage with new test cases and new distros, fixing minor issues on the way

Jakuje:
Tests:
 * Ubuntu and Debian CI targets
 * Backup & Restore for each test case to make sure the tests do not depend on each other
 * Verify OS defaults are not changed when role is executed without arguments
 * Verify variable precedence
 * Test role backup functionality
 * RHEL6 compatibility

Fixes:
 * RHEL8 default configuration
 * Debian default configuration
 * Create runtime directory on Debian based systems (for check in case service is not running)
 * Fix vars precedence to work with Debian versions (before, Debian.yml was loaded instead of Debian_9.yml)
 * Clarify match blocks in README

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -84,6 +84,16 @@
       changed_when: False
       when: sshd_test_hostkey.path is defined
 
+    - name: Make sure sshd runtime directory is present
+      file:
+        path: "{{ __sshd_runtime_directory }}"
+        state: directory
+        owner: root
+        group: root
+        mode: "{{ __sshd_runtime_directory_mode }}"
+      when:
+        - __sshd_runtime_directory | d(false) | bool
+
     - name: Create the configuration file
       template:
         src: sshd_config.j2
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -39,7 +39,11 @@
       {% if not sshd_verify_hostkeys %}
         []
       {% elif sshd_verify_hostkeys == 'auto' %}
-        {{ __sshd_hostkeys_from_config }}
+        {% if sshd_HostKey is string %}
+          [ {{ __sshd_hostkeys_from_config }} ]
+        {% else %}
+          {{ __sshd_hostkeys_from_config }}
+        {% endif %}
       {% else %}
         {{ sshd_verify_hostkeys | to_json }}
       {% endif %}
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -19,6 +19,7 @@
     backup: "{{ sshd_backup }}"
   when:
     - sshd_sysconfig|bool
+    - __sshd_sysconfig_supports_use_strong_rng or __sshd_sysconfig_supports_crypto_policy
   notify: reload_sshd
 
 - name: Make sure hostkeys are available and have expected permissions
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -157,6 +157,7 @@
   when:
     - sshd_manage_service|bool
     - ansible_virtualization_type|default(None) != 'docker'
+    - ansible_virtualization_type|default(None) != 'podman'
     - ansible_virtualization_type|default(None) != 'VirtualPC' # for Github Actions
     - ansible_connection != 'chroot'
 
--- a/tests/tasks/backup.yml
+++ b/tests/tasks/backup.yml
@@ -1,0 +1,25 @@
+---
+- name: Create a temporary directory for backup files
+  tempfile:
+    state: directory
+  register: __sshd_test_backup
+  changed_when: False
+  when:
+    - sshd_test_backup_skip is not defined
+
+- name: Make sure openssh is installed before creating backup
+  package:
+    name: openssh-server
+    state: present
+
+- name: Backup files
+  shell: >
+    if test -f {{ item }}; then
+      mkdir -p {{ __sshd_test_backup.path }}/$(dirname {{ item }});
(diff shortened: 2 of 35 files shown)


tasks/install.yml
```diff
@@ -80,6 +85,17 @@
...
+    - name: Make sure sshd runtime directory is present
+      file:
+        path: "{{ __sshd_runtime_directory }}"
+        state: directory
+        owner: root
+        group: root
+        mode: "{{ __sshd_runtime_directory_mode }}"
+      when:
+        - __sshd_runtime_directory is defined
+        - __sshd_runtime_directory != false
```
richm, line 97: `when: __sshd_runtime_directory | d(false)`
or if it is possible that `__sshd_runtime_directory` does not evaluate to a boolean you could use to be sure
`when: __sshd_runtime_directory | d(false) | bool`

tests/tasks/backup.yml
```diff
@@ -0,0 +1,24 @@
...
+    - sshd_test_backup_skip is not defined
+
+- name: Backup files
+  shell: >
+    if test -f {{ item }}; then
+      mkdir -p {{ __sshd_test_backup.path }}/$(dirname {{ item }});
+      cp {{ item }} {{ __sshd_test_backup.path }}/$(dirname {{ item }})
+    fi
+  changed_when: False
+  loop: "{{ __sshd_test_backup_files }}"
```
richm, line 20: This is one of those gotchas with ansible - the `when` are evaluated for each `loop` iteration - so that means `  loop: "{{ __sshd_test_backup_files }}"` could be evaluated and not be protected by the `__sshd_test_backup_files is defined` - so I suggest using `  loop: "{{ __sshd_test_backup_files | d([]) }}"`

tests/tasks/restore.yml
```diff
@@ -0,0 +1,36 @@
+---
+- name: Restore backed up files and remove what was not present
+  shell: >
+    if test -f {{ __sshd_test_backup.path }}/{{ item }}; then
+      cp {{ __sshd_test_backup.path }}/{{ item }} $(dirname {{ item }})
+    elif test -f {{ item }}; then
+      rm {{ item }}
+    fi
+  changed_when: False
+  loop: "{{ __sshd_test_backup_files }}"
```
richm, line 10: same here - that is, if it is at all possible that `__sshd_test_backup_files` could be undefined

tests/tests_backup.yml
```diff
@@ -0,0 +1,67 @@
...
+    find:
+      paths: "{{ main_sshd_config_path }}"
+      patterns: "{{ main_sshd_config }}.*@*~"
+    register: new_backup
+
+  - name: Verify the backup was not done in the first attempt, but in the second one
+    assert:
+      that:
+        - "no_backup.files == []"
+        - "new_backup.files != []"
```
richm, line 64: you don't need quotes for `assert.that` or `when` conditions - ansible is smart enough to evaluate them as jinja code - one exception is if the condition begins with a quoted string or a {} or a []

tests/tests_hostkeys_missing.yml
```diff
@@ -24,10 +32,26 @@
             - ansible_failed_result.msg != 'UNREACH'
             - not role_result.changed
           msg: "Role has not failed when it should have"
+        when:
+          - ansible_facts['os_family'] != 'Debian'
+          - ansible_facts['distribution'] != 'RedHat' and ansible_facts['distribution_major_version'] != 6
```
richm, line 37: do you mean `- not (ansible_facts['distribution'] == 'RedHat' and ansible_facts['distribution_major_version'] == 6)` ?

tests/tests_os_defaults.yml
```diff
@@ -0,0 +1,34 @@
...
+      name: ansible-sshd
+  - name: Show effective configuration after running role (role defaults)
+    shell: sshd -T
+    register: runtime_after
+  - name: Check that the effective configuration did not change from OS defaults
+    assert:
+      that:
+        - "runtime_before.stdout == runtime_after.stdout"
+    when:
+      - ansible_facts['distribution'] != 'RedHat' and ansible_facts['distribution_major_version'] != 6
```
richm, line 31: see above

tests/tests_hostkeys_missing.yml
Jakuje: Right. Thanks for checking. It turns out the 6 also needs to be in quotes to work correctly.

tests/tasks/backup.yml
Jakuje: I think I read it, but successfully forgot when I wrote this loop. Changed according to your suggestion. Thanks.

tasks/install.yml
Jakuje: I think I had some issues with the filter `bool` as it was converting strings to `false` when they were different strings than 'true'. But here, it looks like it works. Thanks

Jakuje:
@mattwillsher can you have a look if these changes work for you?

mattwillsher:
Looks ok to me - would you be happy to remove the travis ci checks now? They seem unneeded now there are GH actions

Jakuje:
The travis is terribly slow these days. The only test that is not running now in the Github Action is the idempotency test as there is really not a simple way to run it as a simple playbook. The only solution I was advised to use is starting the second run of a ansible-playbook from the first playbook, but I did not attempt to implement it yet. Given that, unless there will be some further issues with it, I think we should keep it before we will have some alternative for this test.

PR #148, opened 2020-11-28
Add explicit become on run

sorrowless:


Jakuje:
I do not think this is a good idea. The whole connection to the server might be already running as a root and then `become` is just spamming audit with sudo events changing root to root.

You can already run this role as a non-root user with become privileges as in the example playbook in the readme.

You are right about the shell part, but please, submit it in separate commit as it does not have anything to do with the rest of the changes.

sorrowless:
> You can already run this role as a non-root user with become privileges as in the example playbook in the readme.

It is possible, yes. But for cases when your role is just a part of bigger playbook and it supposed to run rove over role (e.g. set pam configuration, create users, set sshd, set ssh keys, whatever), you have to move some of these roles to set with 'become: true' and set 'before: false' for these which depends on user identity. In case you have many roles and you have to support some ordering for them, it became a mess pretty fast, as your playbook became look like

```yaml
hosts: servers
become: true
roles:
  - pam

hosts: servers
roles:
  - users
  - ssh_keys

hosts: servers
become: true
roles:
  - sudoers
  - sshd

hosts: servers
roles:
  - postgresql
```

In case all the roles can be ran by ordinary user which have a sudo though (usually such things deployed with special user for deploy only), playbook become to

```yaml
hosts: servers
roles:
  - users
  - ssh_keys
  - sudoers
  - sshd
  - postgresql
```

for cases when you have 15+ roles applied or for cases when you generate your playbook on the fly, second version is a way more clean. That's why I suppose to have explicit 'become' in role itself, not setting it on the playbook side. One can say that you can simply set 'become: true' for all the roles in the playbook but unfortunately it leads to wrong default file owner and permissions and as a result you have to rewrite bunch of roles which you apply to something like

```yaml
- task:
  copy:
    src: filename
    dest: /target/path/
    owner: "{{ ansible_user }}"
    group: "{{ ansible_user }}"
    mode: 0600
```

cause if you won't do that, default owner will be root and you want that almost nowhere. Another problem lies in the fact that 'ansible_user' variable is not always accessible, for example in molecule tests, so you have to add redundant vars just to tests.

Another pick is that looking to the role where tasks are explicitly marked by 'become: true' one can easily understand if that task will really run as root in any case - or fail. Setting it in readme examples is good - but I personally believe that explicit is better than implicit and setting 'become: true' explicitly is for greater good.

I still understand your concerns though, so if you decide to not merge this anyway - feel free to close this pull request, I will just accept these changes in my fork.

And - have a good day and thank you for this role btw. I use it for some time already and it helps me alot.

Jakuje:
Do you have an example of a role that does not need to run as a root or does not need a `become`? Would it make sense to set `become: false` to these roles/tasks that should not run as root or elevated users?

mattwillsher:
I'm not against this change as such. It seems at visual inspection to offer finer grain control of permissions which I'm in favour of.

I have seen environments where root is not allowed to sudo to root, which I'm concerned could cause this new code to produce hard to diagnose errors.

This is not (although perhaps it should be) common practice in Ansible code, to my knowledge.

PR #149, opened 2020-12-07
feat: add support for non-exclusive configuration

jpiron:
This is an updated version of https://github.com/willshersystems/ansible-sshd/pull/71.
The idea is still the same: provide the ability configure SSH in a non exclusive manner.

The main use case is for other playbooks, tasks or roles to update SSH configuration while preserving idempotency:
I have a playbook `common.yml` that configures server basics that uses this role to configure SSH with default values.
When I later add another playbook for a specific software for instance that requires some specific SSH configuration, I want this specific configuration to be part of this software playbook to ensure the playbook contains everything that software requires (kind of SOLID principles applied to playbooks). But I don't want further `common.yml` runs to overwrite it. 

This provides kind of the same functionality as the `sshd_config` [Include](https://bugzilla.mindrot.org/show_bug.cgi?id=2468) that has been added to  OpenSSH 8.1 for older versions.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -95,6 +95,47 @@
           {% endif %}
         backup: "{{ sshd_backup }}"
       notify: reload_sshd
+      when: sshd_exclusive_configuration | bool
+
+    - name: Non-exclusive configuration
+      when: not sshd_exclusive_configuration | bool
+      vars:
+        marker_begin: BEGIN
+        marker: "# {mark} ANSIBLE SSH MANAGED CONFIGURATION"
+      block:
+        - name: Check if configuration block is present
+          lineinfile:
+            path: "{{ sshd_config_file }}"
+            line: "{{ marker.format(mark=marker_begin) }}"
+            state: present
+          check_mode: true
+          register: configuration_block_presence
+
+        - name: Remove default configuration
+          file:
+            path: "{{ sshd_config_file }}"
+            state: absent
+          when: configuration_block_presence is changed
+
+        - name: Copy configuration
+          blockinfile:
+            block: "{{ lookup('template', 'sshd_config.j2') }}"
+            dest: "{{ sshd_config_file }}"
+            create: true
+            owner: "{{ sshd_config_owner }}"
+            group: "{{ sshd_config_group }}"
+            mode: "{{ sshd_config_mode }}"
+            marker_begin: "{{ marker_begin }}"
+            marker: "{{ marker }}"
+            validate: >-
+              {% if sshd_test_hostkey is defined and sshd_test_hostkey.path is defined %}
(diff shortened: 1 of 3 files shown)


Jakuje:
Thank you for the contribution. This was something I wanted to look into as the include functionality is not yet well established. Can you provide also some testcases with this functionality to make sure it works as expected?

tasks/install.yml
```diff
@@ -95,6 +95,47 @@
...
+            mode: "{{ sshd_config_mode }}"
+            marker_begin: "{{ marker_begin }}"
+            marker: "{{ marker }}"
+            validate: >-
+              {% if sshd_test_hostkey is defined and sshd_test_hostkey.path is defined %}
+                {{ sshd_binary }} -t -f %s -h {{ sshd_test_hostkey.path }}/rsa_key
+              {% else %}
+                {{ sshd_binary }} -t -f %s
+              {% endif %}
+            backup: "{{ sshd_backup }}"
```
Jakuje, line 136: If this task fails (for example in the validation), you are left without the configuration file at all as you removed it in the previous task.

tasks/install.yml
```diff
@@ -95,6 +95,47 @@
...
+        marker_begin: BEGIN
+        marker: "# {mark} ANSIBLE SSH MANAGED CONFIGURATION"
+      block:
+        - name: Check if configuration block is present
+          lineinfile:
+            path: "{{ sshd_config_file }}"
+            line: "{{ marker.format(mark=marker_begin) }}"
+            state: present
+          check_mode: true
+          register: configuration_block_presence
```
Jakuje, line 112: This checks for the presence of the beginning marker and if it is not there, it adds it? I am wondering if these two steps are needed at all, whether it should not be enough to use the `blockinfile`.

I am also wondering whether the blocks should be named (through some other variable). In case, you will have `sftp` role, which will add just a `Subsystem sftp internal-sftp` in first version, but in second version, it adds some more options and you run it again (with `sshd_exclusive_configuration: false`), it will create second block and fail as it is not valid configuration to have two `sftp` subsystems (even if they are same).

Using `sshd_exclusive_configuration: true` on default configuration file will remove the file so no defaults are preserved (not OS defaults nor defaults set by this role with `sshd_exclusive_configuration: false`).

Jakuje (commented):
I think it needs some work and considerations, see comments

jpiron:
Indeed, it requires some additional work, I'll look a this.
Thanks for your feedback

Jakuje:
Can you check if something like #158 would work for you?

Jakuje:
The alternative was implemented and is now release with 0.13.0 version. If you will see some issues with it, please open a new issue or PR.

PR #152, opened 2021-01-06
README: Fix variable name

Jakuje:


Jakuje:
I think this one is already in as 9fba3f579431748d1999d7ed9f3726728244bc92 from #150.

PR #153, opened 2021-01-27
[FreeBSD] Add Subsystem to _sshd_defaults

Migsi:
The Subsystem entry was missing for FreeBSD OS, noticed this while provisioning a TrueNAS box. After the first provision ansible was unable to upload any files due to that missing setting. Tested this change by adjusting the role locally and rerunning it with a clean sshd_config on the remote side, worked fine.


Jakuje (approved):

Jakuje:
Thank you! Merged.

Issue #94, opened 2018-09-11
Append sshd config


beoSheep:
Hi,

Is it possible to Append some part of sshd_config file? My case is that I need to add some values in **AcceptEnv**, but it overwrites. Any suggestion?

tterranigma:
There is no build in support for appending. There is [an old PR](https://github.com/willshersystems/ansible-sshd/pull/71) that you could perhaps edit.

beoSheep:
Huh, I am not that experienced with Ansible yet. My use case is, that I have some roles which have additional **AcceptEnv** lines comparing to global sshd config, so for me it's a duplicate configuration or maybe I am doing it wrong.. :)

mattwillsher:
AcceptEnv would need to be duplicated in your code at the moment.

beoSheep:
Is there any plans to add such feature?

mattwillsher:
I've been considering how such a feature could be added, but no plans at the moment. If I come up with a clean way to do it I will add it then.

tterranigma:
@mattwillsher Personally, I can think of only two ways this could be implemented. One is what that old PR did.

The other is to have two extra variables (the names are on the spot):

- `sshd_config_append_only`: False by default, if set, the role will only append configuration to the config file.
- `sshd_config_block`: Empty by default, it contains the config snippet to append to the config file.

The current template could also use `sshd_config_block` and print eg at the end of the file. 

So, there would be 3 possibilities:

- config template only
- config template plus config block
- config block only

Jakuje:
With OpenSSH 8.3p1 (in Fedora 32, probably others), this can be done by writing a file in drop-in directory (`/etc/ssh/sshd_config.d/`) as openssh server config supports `Include` keyword.

mimmus:
Appending only, you give up to effectively managing your service. In other words, you haven't a "state" to apply to a service, to be sure that it has a fixed behaviour.
In any case, I agree with you that effectively overwriting the default config file is not always a a desirable behavior.
Solutions are:
- modifying inline the original file (using a module as 'lineinfile'
- dropping-in a file under /etc/yourservice.d directory, if available (as suggested above)

My 2c, even in the bad place (sorry)

Jakuje:
Can you check if something like #158 would work for you?

mimmus:
> Can you check if something like #158 would work for you?
Does this preserve original sshd_config file?
[...]

PR #158, opened 2021-05-20
Support for appending a snippet to configuration file

Jakuje:
This should address #94 and is alternative approach to #149.

This allows appending non-conflicting configuration snippets to the main configuration file for platforms that do not support drop-in directory. The results should be idempotent and (for non-conflicting configurations) same regardless the order of applying. For conflicting options, standard limitations apply as only the first found option is effective.


Jakuje:
This change is by no means final, but it should cover most of the use cases. I can think of one more addition to prepend a snippet, mostly because of limitations of the sshd configuration parser, to make sure some snippet is coming before any others, but reviews and comments welcomed.

Same from @richm and @mattwillsher

README.md
```diff
@@ -156,6 +156,23 @@ if the system does not have hardware random number generator.
...
+* `sshd_namespace_append`
+
+By default (*null*), the role defines whole content of the configuration file
+(with potential system defaults). To allow this role to be invoked from other
+roles or from multiple places in a single playbok on systems that do not
+support drop-in directory, we can define namespaces, that will allow us place
+configuration snippets idempotently into a single configuration file. The only
+requirement for these instances is to have different namespace name (this
+variable). Other limitation of openssh configuration file such as that only the
+first option specified in a configuration file is effective still apply.
```
richm, line 168: Take a look at https://linux-system-roles.github.io/documentation/incremental_settings.html for an alternate approach

Jakuje: Thank you for the link. This approach would require a lot of new code including the whole sshd configuration parser in this role (because we can not just use the `sshd` as it evaluates match blocks which we need to preserve). It would also require (at least slight) changes of the API.

README.md
```diff
@@ -156,6 +156,23 @@ if the system does not have hardware random number generator.
...
+
+Technially, the snippets are placed in `Match all` blocks (unless they contain
```
richm, line 170: Technically

README.md
richm: ok - then your approach looks fine

templates/sshd_config_snippet.j2
```diff
@@ -0,0 +1,241 @@
+{% macro render_option(key,value,indent=true) %}
+{%   if value is defined %}
+{%     if indent == true %}  {% endif %}
```
richm, line 3: `if indent` - since `indent` is a boolean, you don't need to compare it to `true`

templates/sshd_config_snippet.j2
```diff
@@ -0,0 +1,241 @@
...
+{%     endif %}
+{%   endif %}
+{% endmacro %}
+{% macro body_option(key,override) %}
+{%   set value = undefined %}
+{%   if override is defined %}
+{%     set value = override %}
+{%   elif sshd[key] is defined %}
+{%     set value = sshd[key] %}
+{%   elif __sshd_defaults[key] is defined and sshd_skip_defaults != true %}
```
richm, line 23: `and not sshd_skip_defaults` (assuming `sshd_skip_defaults` is boolean valued)

tests/tests_namespace_append.yml
```diff
@@ -0,0 +1,74 @@
...
+      vars:
+        sshd_config_file: /etc/ssh/sshd_config
+        sshd_skip_defaults: true
+        sshd_namespace_append: nm2
+        sshd:
+          AcceptEnv: LS_COLORS
+          PasswordAuthentication: no
+          Match:
+            Condition: Address 127.0.0.1
+            AllowTcpForwarding: no
```
richm, line 36: Can you give an example of what the sshd config will look like at this point?

templates/sshd_config_snippet.j2
Jakuje: this file is generated from others in `meta/`, but I can fix it anyway.

README.md
Jakuje: Thanks. Fixed

tests/tests_namespace_append.yml
```diff
@@ -0,0 +1,74 @@
...
+              - "'Match Address 127.0.0.1' in config.content | b64decode"
+              - "'AllowTcpForwarding no' in config.content | b64decode"
+
+        - name: Check the configuration values are effective
+          # note, the options are in lower-case here
+          assert:
+            that:
+              - "'acceptenv EDITOR' in runtime.stdout"
+              - "'acceptenv LS_COLORS' in runtime.stdout"
+              - "'passwordauthentication yes' in runtime.stdout"
```
richm, line 69: Ideally, the test would also check if the given string is in the given match block e.g. it would check that `AcceptEnv EDITOR` is in one of the `Match all` blocks; it would check that `AllowAgentForwarding no` is in the `Match user root` block; etc.  Is this possible? e.g. using https://docs.ansible.com/ansible/2.8/user_guide/playbooks_tests.html#testing-strings

Jakuje: The `sshd -T` evaluates the match blocks so at this moment it checks that both of the checks are evaluated positively and reflected in the output.

I should probably add also negative test (another `sshd -T` invocation with different parameters) to check that the content of match blocks is not reflected in these. Should not be that hard.

Testing the "context" is probably also a good idea. Let me try.

richm (approved):
lgtm

mattwillsher (approved):

mattwillsher:
Also looks fine to me. I want to look at that travis check and remove it if it's no needed anymore. Then you should be able to merge and tag and get the a new release out I think @Jakuje

Jakuje:
Thank you for the reviews. Lets keep it open for some days. I hope I will be able to get back to that, test it more and write some prove of concept using this.

README.md
```diff
@@ -156,6 +156,23 @@ if the system does not have hardware random number generator.
 The path where the openssh configuration produced by this role should be saved.
 This is useful mostly when generating configuration snippets to Include.
 
+* `sshd_namespace_append`
+
+By default (*null*), the role defines whole content of the configuration file
```
spetrosi, line 161: I think that an example of using this variable will be beneficial, e.g. in some example playbook.

README.md
```diff
@@ -156,6 +156,23 @@ if the system does not have hardware random number generator.
...
+support drop-in directory, we can define namespaces, that will allow us place
+configuration snippets idempotently into a single configuration file. The only
+requirement for these instances is to have different namespace name (this
+variable). Other limitation of openssh configuration file such as that only the
+first option specified in a configuration file is effective still apply.
+
+Technically, the snippets are placed in `Match all` blocks (unless they contain
+other match block) to make sure they are applied regardless the previous match
+blocks. This allows to configure any non-conflicting options from different
+roles invocations.
```
spetrosi, line 173: Trying to simplify and clarify the description a bit:

By default (*null*), the role defines the whole content of the configuration file with
potential system defaults. You can use this variable to invoke this role from other
roles or from multiple places in a single playbook on systems that do not support
drop-in directory.

When this variable is set, the role places the configuration that you apply with the
`sshd` variable to configuration snippets in a single configuration file under the set
namespace name. You must set this variable to a different namespace name when
invoking the role several times.                                                             <-- Please check if this sentence is true

Note that limitations of the openssh configuration file, such as that only the
first option specified in a configuration file is effective, still apply.

Technically, the role places the snippets in `Match all` blocks unless they contain
other match blocks to ensure that they are applied regardless of the previous match
blocks. This allows configuring any non-conflicting options from different
roles invocations.

README.md
spetrosi: As well as an example resulting config file, just to give a wider overview of what this variable does.

Jakuje:
Thank you for the suggestion @spetrosi. I tried to improve the wording according to your suggestions. Please, check if it looks fine.

There are couple of ideas that might make sense or might make the feature more complicated:
 * I think this option should automatically turn on `sshd_skip_defaults` as for snippets. It does not make sense to have the os defaults again in every snippet (done in current version)
 * For newer systems that support drop-in directory (Fedora), it would make sense to make this variable to fall back to change the `sshd_config_file` to something like `/etc/ssh/sshd_config.d/{namespace}.conf` similarly as we do in ssh role.

This would simplify configuring mix of different system versions.

README.md
Jakuje: added both

Jakuje:
@mattwillsher we discussed last week the naming of this configuration option and one of better suggestions was `sshd_config_namespace` as the appending is not completely describing what the option does. What do you think?

As we did not yet release a new version, it might make sense to change it rather sooner than later after we will have a release. If you agree, I can submit a new PR before we will do a release.

Jakuje:
Thanks. Submitted #160. I will check why it failed in Fedora today.

PR #159, opened 2021-06-02
Fix variable precedence when invoked through legacy "roles:"

Jakuje:
While the variable precedence (for example for `sshd_hostkey_mode` variable) works fine when the role is invoked through `include_role:`, it does not when it is invoked through the `roles:`. This is caused by complicated variable precedence. The file `vars/RedHat_8.yml` is loaded (and it gets precedence 18), but it is supposed to be possible to override with the variables from the playbook. When invoking the role through `include_role`, it gets precedence 20, which is fine, but it looks like plain vars in play (when invoked through `roles:`) have precedence only 12, which is probably the cause.

This PR is changing the role internals so these variables are possible to override from plays in either of the invocations, adds tests for these cases and adjusts the documentation to mark these variables as the role API (not just secondary ones).

[1] https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable


richm (approved):
lgtm

richm:
@Jakuje ready to merge?

Jakuje:
I would like @mattwillsher to have a look, as it is quite large change, but if not, I can merge it in coming days.

mattwillsher:
This looks good to me. Please feel free to merge and tag.

PR #160, opened 2021-06-10
Rename sshd_namespace_append to sshd_config_namespace

Jakuje:
The new name better explains what the option does.


Jakuje:
@mattwillsher @richm I think this is the final version. Please, check if that looks fine and the wording after renaming looks reasonable. I had to update also one test that was failing in recent Fedora in CI.

richm (approved):
lgtm

mattwillsher (approved):
👍🏼

PR #161, opened 2021-06-11
Add configuration options from OpenSSH 8.6p1

Jakuje:
@mattwillsher hopefully one last. I remembered that there was some revamp of options in OpenSSH 8.6p1 so I updated them also here. That should be the last thing before the release ;)


Jakuje:
There is a draft release in https://github.com/willshersystems/ansible-sshd/releases -- hopefully it will be visible also for you. I would welcome a review if I missed something important.

mattwillsher (approved):

PR #163, opened 2021-08-03
Add support for RHEL 9 and adjust tests for it

Jakuje:
This adds a proper support for RHEL9 (based on latest Fedora) and adjusts test for that.

### DIFF
--- a/tests/tasks/setup.yml
+++ b/tests/tasks/setup.yml
@@ -13,3 +13,19 @@
     mode: "0755"
   when:
     - ansible_facts['os_family'] == 'Debian'
+
+- name: Define common variables
+  set_fact:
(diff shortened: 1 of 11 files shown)


Jakuje:
@mattwillsher The travis-ci is no longer working, but it is still listed here as a required check preventing merging. I think you need to remove it somewhere in settings? We can then remove also the `.travis.yml`

@richm @nhosoi I would like to ask you for your review here.

nhosoi (approved):
lgtm
Thanks for the quick fix, @Jakuje!

mattwillsher:
@Jakuje I've removed travis from the required checks.

mattwillsher (approved):

Jakuje:
Thanks!

PR #138, opened 2020-09-24
Add linting for prior version back to Actions

mattwillsher:
Can be merged once upstream linting actions work again.

PR #164, opened 2021-08-06
Drop support for Ansible 2.8 by bumping the Ansible version to 2.9

spetrosi:
Bug 1989197 - drop support for Ansible 2.8
https://bugzilla.redhat.com/show_bug.cgi?id=1989197

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -13,7 +13,7 @@ jobs:
   #       with:
   #         targets: "tests/test_*.yml"
   #         override-deps: |
-  #           ansible==2.8
+  #           ansible==2.9
   #         args: ""
   # test-ansible29:
   #   runs-on: ubuntu-latest
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -4,7 +4,7 @@ galaxy_info:
   description: OpenSSH SSH daemon configuration
   company: Willsher Systems
   license: LGPLv3
-  min_ansible_version: 2.8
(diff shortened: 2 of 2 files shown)


Jakuje:
we just recently bumpted the minimum ansible version to 2.8 in (#127), but I am not sure if we need to support 2.8 anymore. @mattwillsher what do you think?

mattwillsher:
I don't see the need to drop support unless there is functionality we require that 2.8 doesn't support.

spetrosi:
@mattwillsher [linux-system-roles](https://github.com/linux-system-roles) decided to drop support for Ansible 2.8 because it reached its EOL on January 21, 2021, and some roles faced some errors related to 2.8 version. Hence, linux-system-roles remove tests for Ansible 2.8 from the CI, meaning that there is no guarantee that roles work on Ansible 2.8. Even though some roles might work on Ansible 2.8, keeping the support for a version that is EOL is too much effort with less usefulness.

mattwillsher:
Ok, I'm happy to go with the consensus on this if that's the community standard.

mattwillsher:
It's arguable that min rev is really 2.10 #138 

I'll try to spend a bit of time tidying this up generally. 

I don't see an immediate need disbar 2.8,  and it's a good chance to do house keeping and properly have a tested minimum version.

spetrosi:
@mattwillsher Can we merge this PR?

mattwillsher:
Yes it will be merged, but I want to get a release out for CentOS 6 and possible Debian 11 before this goes in.

PR #165, opened 2021-08-09
Add CentOS 6 to CI

Jakuje:
The RHEL6/CentOS 6 has a lot of issues, but we still need to make sure it works at least with best effort.

This PR removes one incompatibility which was introduced with the namespaces, adjusts testsuite to work there and adds Github Action for CentOS6 (with a tweak removing the `end_host` part, which is not compatible with Ansible version available there).


richm (approved):
lgtm

mattwillsher:
I'll release this and if ready the Debian 11 PR before cutting off Ansible 2.8.

PR #166, opened 2021-08-09
Add Debian 11 (bullseye) support

domrim:
This PR adds bullseye support to the role.
The workflow fails at the moment because there is no `roles-ansible/check-ansible-debian-bullseye-action` repo yet.
Debian 11 will be released at the 14. Aug. So it would be nice to have this role ready till saturday.

### DIFF
--- a/.github/workflows/ansible-debian-bullseye.yml
+++ b/.github/workflows/ansible-debian-bullseye.yml
@@ -1,0 +1,18 @@
+name: Run tests on Debian bullseye (11)
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      # Important: This sets up your GITHUB_WORKSPACE environment variable
+      - name: checkout PR
+        uses: actions/checkout@v2
+
+      - name: ansible check with debian:bullseye (11)
+        uses: roles-ansible/check-ansible-debian-bullseye-action@master
+        with:
+          group: local
+          hosts: localhost
+          targets: "tests/*.yml"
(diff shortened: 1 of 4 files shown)


Jakuje (approved):
looks good. Lets see if the actions will work. If so, I will merge it and add a new release.

Jakuje:
> Failed to resolve action download info. Error: Unable to resolve action `roles-ansible/check-ansible-debian-bullseye-action@master`, repository not found

yeah .. the github action for latest debian does not exist yet ...

Jakuje:
Possible workaround might be trying the https://github.com/roles-ansible/check-ansible-debian-latest-action (not sure if it already does bullseye) or asking @DO1JLR to add the action.

DO1JLR:
By the way, the https://github.com/marketplace/actions/check-ansible-debian-bullseye is now available.

Sorry for the delay :wink:

DO1JLR:
The check-ansible-debian-bullseye gh-action is available at the 'main' branch btw.
https://github.com/domrim/ansible-sshd/pull/1

Jakuje:
@DO1JLR thanks for prompt reply and creating the new action! 
@domrim it looks like the debian-latest also fails now with the following error:
```
  assertion: runtime_before.stdout == runtime_after.stdout
```
can you check what are the defaults of the new debian and update the PR?

domrim:
okay, i'll look into it

domrim:
> @DO1JLR thanks for prompt reply and creating the new action!
> @domrim it looks like the debian-latest also fails now with the following error:
> 
> ```
>   assertion: runtime_before.stdout == runtime_after.stdout
> ```
> 
> can you check what are the defaults of the new debian and update the PR?

after running the actions again, everything seems fine. also debian-latest & debian-bullseye checks should be the same. don't know why it failed before

Jakuje:
Oh, right. Probably it fall backs to some defaults, before we had your `vars/Debian_11.yml` in place. Thanks. Looks good now.

PR #169, opened 2021-08-17
Remove travis configuration and update readme with new badges

Jakuje:


Jakuje:
The only failing job is now debian-latest, which should be handled with #166.

Jakuje:
@mattwillsher I think this is last thing that should go in before release.

PR #178, opened 2022-04-06
Make sure Include is in the main configuration file when drop-in directory is used

Jakuje:
This implements sanity check to make sure the configuration files on RHEL9 and Fedora (and now also Ubuntu 22.04), which are placed in drop-in directory are really applied by checking the include directive is present in the main configuration file.

There are also CI fixes, refactoring and the whole new support for new Ubuntu 22.04.

PR #179, opened 2022-04-11
Unbreak FIPS detection and stabilize failing tests and GH actions

Jakuje:
Subset of bugfix changes from #178


tasks/install.yml
```diff
@@ -41,9 +41,9 @@
 - name: Make sure hostkeys are available and have expected permissions
   vars: &share_vars
     __sshd_fips_mode: >-
-      - __sshd_hostkeys_nofips | d([])
-      - __sshd_kernel_fips_mode.content | b64decode == "1" | bool or \
-        __sshd_userspace_fips_mode.content | b64decode != "0" | bool
+      {{ __sshd_hostkeys_nofips | d([]) and
+         (__sshd_kernel_fips_mode.content | d('MAo=') | b64decode | trim == '1' or
```
mattwillsher, line 45: What's this base 64 string?

mattwillsher (approved):
Look fine to me, just a question about the base64 string

tasks/install.yml
richm:
```
python
>>> import base64
>>> ss = 'MAo='
>>> base64.b64decode(ss)
b'0\n'
```
e.g. the default value for fips mode is `0`

tests/tests_alternative_file.yml
```diff
@@ -80,7 +82,7 @@
 
         - name: Print the main configuration file
           slurp:
-            src: "{{ main_sshd_config }}"
```
richm, line 83: What is the motivation for this change?

tasks/install.yml
Jakuje: Right. Just taken from some local testing. Not sure if there is a nicer way to write these.

richm: I think this is fine - maybe add a comment like `# 'MAo=' evaluates to '0'`

tests/tests_alternative_file.yml
Jakuje: Its from a99f9e1fd5b549da6a911d71e4b4d59c57bce0d4. The test `tests/tests_alternative_file.yml` writes several configuration files, and checks that options are set in the respective ones. The variable `main_sshd_config` is different if the system supports drop-in directory or not so I am not sure how this ever worked before. On line 52 we force writing to `/etc/ssh/sshd_config` so this change is syncing on that.

tests/tasks/restore.yml
```diff
@@ -30,11 +30,16 @@
     name: sshd
     state: reloaded
   changed_when: false
+  vars:
+    __sshd_skip_virt_env:
```
richm, line 34: This is defined at https://github.com/willshersystems/ansible-sshd/pull/179/files#diff-23e21b6c89468f7534697ad091835b3ee5e0213b37ace489488234e5a9548ec4R85 ?

You could use the list variable from the role, but you would have to refactor the role somewhat - see https://github.com/redhat-cop/automation-good-practices/pull/62

tests/tests_hostkeys_fips.yml
```diff
@@ -4,13 +4,52 @@
...
+          stat:
+            path: /etc/ssh/ssh_host_ed25519_key.pub
+          register: pubkey
+
+        - name: Check the key is in configuration file (without include)
+          assert:
+            that:
+              - "'HostKey /etc/ssh/ssh_host_ed25519_key' in config.content | b64decode"
+          when:
+            - ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version']|int < 9
```
richm, line 42:
```yaml
          when:
            - ansible_facts['os_family'] == 'RedHat'
            - ansible_facts['distribution_major_version']|int < 9
```

tests/tests_hostkeys_fips.yml
```diff
@@ -4,13 +4,52 @@
...
+
+        - name: Check host key was generated
+          assert:
+            that:
+              - privkey.stat.exists
+              - pubkey.stat.exists
+      when:
+        - ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version']|int > 6
```
richm, line 50:
```yaml
      when:
        - ansible_facts['os_family'] == 'RedHat'
        - ansible_facts['distribution_major_version']|int > 6
```

tests/tests_hostkeys_fips.yml
Jakuje: Thanks. Changed that also in the other places where this was present.

tests/tasks/restore.yml
Jakuje: I factored out this variable to separate file and it looks like it works ok in my case.

Not sure how the setup part is related now. Should I add it too? How common is it to have the `gather_facts: false`?

richm: > I factored out this variable to separate file and it looks like it works ok in my case.
> 
> Not sure how the setup part is related now. Should I add it too? How common is it to have the `gather_facts: false`?

It is common enough that we are fixing all of the system roles to handle this case - so I suggest you add support for this to sshd also.

tests/tasks/restore.yml
```diff
@@ -25,16 +25,16 @@
     - __sshd_test_backup is defined
     - __sshd_test_backup.path is defined
 
+- name: Include common variables
+  include_vars: ../../vars/common.yml
```
richm, line 29: This won't work when doing testing of the converted collection - some of our tests specifically move the `tests/` directory out of the collection so that we can all of the collection path lookups work correctly (i.e. when you run a playbook inside a collection some of the path lookups work implicitly - but an end user playbook will be outside of the collection).
This is why it is recommended to use the idiom of https://github.com/redhat-cop/automation-good-practices/pull/62/files#diff-9ca5a441d83a5cc9bff7871d160b57817ffd2d898c420b0aa48165ad444b19c1R188 in the role so that the test can use https://github.com/redhat-cop/automation-good-practices/pull/62/files#diff-9ca5a441d83a5cc9bff7871d160b57817ffd2d898c420b0aa48165ad444b19c1R279

richm (approved):
lgtm

Jakuje:
Thank you for reviews. I filled roles-ansible/check-ansible-ubuntu-latest-action#2 for the failing ubuntu action to see if we can get it green before merging (It was working two days ago).

Unfortunately, it was a bit larger than I initially expected. In any case, I will keep it open for some days before merging if some other review will pop up.

tests/tasks/restore.yml
Jakuje: Thank you. Understood now.

Issue #183, opened 2022-05-12
add IdentityFile to options_body


negast:
Hello

I would like to add an additional private key file to look for on systems as users will have on for connecting and one for git/developer stuff. 

I've found following param defines these paths

> IdentityFile
             Specifies a file from which the user's DSA, ECDSA,
             authenticator-hosted ECDSA, Ed25519, authenticator-hosted
             Ed25519 or RSA authentication identity is read.  The
             default is ~/.ssh/id_dsa, ~/.ssh/id_ecdsa,
             ~/.ssh/id_ecdsa_sk, ~/.ssh/id_ed25519, ~/.ssh/id_ed25519_sk
             and ~/.ssh/id_rsa.  Additionally, any identities
             represented by the authentication agent will be used for
             authentication unless IdentitiesOnly is set.  If no
             certificates have been explicitly specified by
             CertificateFile, ssh(1) will try to load certificate
             information from the filename obtained by appending
             -cert.pub to the path of a specified IdentityFile.

But this is misisng from the options_body and also from the template. Could this be added?

PR #184, opened 2022-05-12
added IndentityFile issue #183

negast:
See issue #183 for reference


Jakuje:
Sorry, there is no such option in the server configuration. For managing client configuration files, please use https://github.com/linux-system-roles/ssh

Issue #170, opened 2021-09-15
sshd_packages unlikely to be overridable


shk3bq4d:
README.md says sshd_packages is free to be overridden. but I don't think the current implementation allows that as the include_vars will unconditionally load one file that would override user selection

maybe a fix would include the use of sshd_skip_defaults ?

Jakuje:
The `sshd_packages` and others are listed under "Secondary role variables". You are able to override it with role arguments if used as within `include_role`, but because of the variable precedence limitation in ansible, its not easy to override them when used with "old" "roles:".

But this variable should be quite static and considered private in the first place, as it is based on the OS. If you need to install some add additional package, it might make sense to provide `sshd_additional_packages` which will be added to the installed list and which should be trivial to add.

nkakouros:
This is fixed in #199.

Jakuje:
This was fixed with #199 and it will be in the next release.

PR #189, opened 2022-06-02
Addition notes about secondary variables

mattwillsher:
Add clarification from #170 to README

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,6 +1,4 @@
 warn_list:  # or 'skip_list' to silence them completely
-  - '106'  # Role name {} does not match ``^[a-z][a-z0-9_]+$`` pattern
   - '306'  # Shells that use pipes should set the pipefail option
-  - 'fqcn-builtins'  # this is not compatible with ansible 2.6 on RHEL6
 exclude_paths:
   - tests/roles/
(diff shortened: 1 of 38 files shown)


mattwillsher:
I'll use this to also fix the linting.

mattwillsher:
Why ignore them? A separate change is worth adding for them perhaps, and bump the minimum version?

mattwillsher:
I think that schema error was because name was missing from the meta/main.yml file. First batch of linting issues have been updated, will finish soon.

Jakuje:
Thanks for having a look into that. I think the centos6 ansible version is too old it does not support fqcn so it will probably require a bit more pre-processing as we already have for other missing features:

https://github.com/willshersystems/ansible-sshd/blob/master/.github/workflows/ansible-centos-check.yml#L11

I would just do something like `sed -i -e 's/ansible.builtin.//g/' */*.yml `

We can also remove the fqcn exception from 

https://github.com/willshersystems/ansible-sshd/blob/master/.ansible-lint#L4

mattwillsher:
CentOS/EL 6 has been EOL for some time now. Should we continue to support it at all?

Jakuje:
> CentOS/EL 6 has been EOL for some time now. Should we continue to support it at all?

CentOS 6 is EOL, but RHEL6 is still in Extended life cycle support (ELS) for at least two years:

https://access.redhat.com/support/policy/updates/errata#Life_Cycle_Dates

We do not support running the ansible controller on RHEL6, but we still want to support users who want to control the legacy nodes. Ideally, we should test the ansible from supported container to centos6, but that does not look like possible with the current github action so I started with the patching which should bring no functional difference.

mattwillsher:
Change is ready for review.

Jakuje (approved):
looks good. Thanks!

PR #190, opened 2022-06-10
tests: Do not be picky about spaces/tabs

Jakuje:
When testing with cloud-init, it modifies the sshd_configuration and can
replace some tabs with whitespaces. This happens frequently around the
subsystem keyword. There are no functional changes, but the matching
did not work as expected.


mattwillsher (approved):
LGTM

richm (approved):
lgtm - works on rhel 9.0 and rhel 8.7

Jakuje:
Thank you for review!

Jakuje:
@mattwillsher I created a draft release in https://github.com/willshersystems/ansible-sshd/releases/tag/untagged-86cd2926ad1060066ba0 -- can you check it looks ok and push it?

mattwillsher:
I wasn't going to tag a release for this as there are no actual code changes that affect end users. 
Could do with a minor rev bump to 1.6.0 as the code is no longer compatible with Ansible v2.9

richm:
> Could do with a minor rev bump to 1.6.0 as the code is no longer compatible with Ansible v2.9

How so?  We have to maintain 2.9 compatibility for our users.

Jakuje:
Oh, what was the 2.9 incompatibility?

mattwillsher:
Update to FQCNs is 2.10, unless I've misunderstood.

mattwillsher:
Maybe I have misunderstood - I see reference to FQCN in a 2.9 guide.

mattwillsher:
Meta/main.yml have the version updated so if 2.9'is ok that will need changing back

PR #191, opened 2022-06-14
the role still supports 2.9

richm:


richm:
The role still works on 2.9, and Red Hat is committed to ensuring the role works on 2.9 until the 2.9 EOL (late 2023)

richm:
@Jakuje fyi

Jakuje:
@richm thanks. Do you have some simple way how to test against the ansible 2.9? I have couple of ideas:

 * I can probably tweak either of the ansible-check roles to install some older version (from pip?) on Fedora
 * Adjust the CentOS 7 role to install ansible from epel rpm instead of pip

But all that requires changes to existing actions. I did not see any ready-made around here.

Jakuje:
oh, actually centos7 action runs now with ansible 2.9 so we should be good:

https://github.com/willshersystems/ansible-sshd/runs/6881575858?check_suite_focus=true

Jakuje (approved):

mattwillsher:
I don't intend to produce a new version as this is purely internal work, but next release will be v1.6.0 due to the amount of changes done here. 

Have to take other views on this though if you this v1.6.0 should go out now.

richm:
> I don't intend to produce a new version as this is purely internal work, but next release will be v1.6.0 due to the amount of changes done here.
> 
> Have to take other views on this though if you this v1.6.0 should go out now.

No, we don't need an official release containing this fix.

Issue #188, opened 2022-05-30
Unexpected templating type error occurred on (<template_content>) TypeError: sequence item 2: expected str instance, int found


simonsunx:
Hi.

I've got error
> Unexpected templating type error occurred on (<template_content>) TypeError: sequence item 2: expected str instance, int found

 while running `ansible-playbook`.

So I changed var types:

- before:
```yaml
sshd:
  PermitRootLogin: no
  ClientAliveCountMax: 99999
  ClientAliveInterval: 20
  RSAAuthentication: yes
  PasswordAuthentication: no
...
```

- after:
```yaml
sshd:
  PermitRootLogin: 'no'
  ClientAliveCountMax: '99999'
  ClientAliveInterval: '20'
  RSAAuthentication: 'yes'
  PasswordAuthentication: 'no'
...
```

After editing error has gone.

Ansible version:
```bash
$ ansible --version
ansible [core 2.13.0]
  config file = /etc/ansible.cfg
  configured module search path = ['/home/login/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/login/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.4 (main, Mar 23 2022, 23:05:40) [GCC 11.2.0]
  jinja version = 3.0.3
  libyaml = True
```

I hope this experience will help someone.

Jakuje:
Thank you for the report. Could this be something new in the ansible 2.13? We are running various ansible versions for some time (currently 2.12 in github actions) and they do not demonstrate this issue.

richm:
> Thank you for the report. Could this be something new in the ansible 2.13? We are running various ansible versions for some time (currently 2.12 in github actions) and they do not demonstrate this issue.

Yep - sounds like an issue with ansible-core 2.13

Jakuje:
Can you clarify if this is an issue *of* ansible-core 2.13 or *of* our role with ansilbe-core-2.13?

richm:
@nhosoi This looks like a 2.13 porting issue - might be an issue with jinja 3 as well - maybe ansible-core 2.13 or jinja 3 is more type strict?  I'm guessing we may have to do some sort of variable type coercion in the template code, but I'm not sure where.

richm:
[...]

PR #195, opened 2022-07-28
Version v0.16.1 - CHANGELOG.md [citest skip]

richm:
[v0.16.1] - 2022-07-28
--------------------

### New Features

- add parameter RSAMinSize

Add support for the new RSAMinSize parameter.

### Bug Fixes

- none

### Other Changes

- Addition notes about secondary variables
- Fix various linting issues
- Revert incorrect module name
- tests: Do not be picky about spaces/tabs

When testing with cloud-init, it modifies the sshd_configuration and can
replace some tabs with whitespaces. This happens frequently around the
subsystem keyword. There are no functional changes, but the matching
did not work as expected.

Signed-off-by: Jakub Jelen <jjelen@redhat.com>

- the role still supports ansible 2.9

- Add CHANGELOG.md

- Add changelog_to_tag.yml to .github/workflows

Description:
When a new changelog section is added to CHANGELOG.md and pushed,
changelog_to_tag.yml is triggered, which generates a new tag and
a new release.

Signed-off-by: Rich Megginson <rmeggins@redhat.com>

PR #197, opened 2022-08-11
Ensure values are cast to correct type

richm:
https://github.com/willshersystems/ansible-sshd/issues/188
Ensure values are cast to string using `| string`
This shouldn't be necessary, but there seems no way to
guarantee using a version of Jinja which doesn't have this
problem.

In addition - it is not good practice to compare values to
`true` or `false` - instead, just ensure the value is a `bool`
type and evaluate in a boolean context.

### DIFF
--- a/templates/sshd_config.j2
+++ b/templates/sshd_config.j2
@@ -7,10 +7,10 @@
 {%     elif value is sameas false %}
 {{ key }} no
 {%     elif value is string or value is number %}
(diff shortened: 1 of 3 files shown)


Jakuje (approved):

richm:
Can this be merged?  then I'll need to update https://github.com/willshersystems/ansible-sshd/pull/195

PR #198, opened 2022-08-16
Fix ansible-lint warnings and work-around false positive

Jakuje:

PR #199, opened 2022-08-16
Fixes un-overrideable public api variables

nkakouros:
This reverts commit f32003f05101f9f81d14c64e4318ed27912fa4ea.

That commit hardcoded many variables to the values under `vars/`, making it impossible for the user to parameterize things like the systemd service name. The assumption was that the `__sshd_*` variables were useless in an effort to blindly adhere to best practices, but they were crucial in allowing flexibility to the user. 

The way the test suite is set up is a bit strange for me, but I tried to make it pass with minimum changes.


nkakouros:
I cannot fix the failing ansible-lint test. The master branch has the same error, so I guess this PR is ready to review.

richm:
This breaks some integration tests - the handler that starts/restarts sshd fails because `sshd_service` is undefined.
```
RUNNING HANDLER [ansible-sshd : Reload the SSH service] ************************
task path: /home/rmeggins/ansible-sshd/handlers/main.yml:3
Tuesday 16 August 2022  10:51:57 -0600 (0:00:00.000)       0:00:24.908 ********
fatal: [/home/rmeggins/.cache/linux-system-roles/centos-9.qcow2]: FAILED! => {
    "changed": false
}

MSG:

missing parameter(s) required by 'state': name
```
This is because the test resets `sshd_service`:
```
TASK [ansible-sshd : Unset variables to allow running the role more than once] ***
task path: /home/rmeggins/ansible-sshd/tasks/sshd.yml:11
Tuesday 16 August 2022  10:51:57 -0600 (0:00:00.030)       0:00:24.788 ********
...
ok: [/home/rmeggins/.cache/linux-system-roles/centos-9.qcow2] => (item=sshd_service) => {
    "ansible_facts": {
        "sshd_service": null
    },
    "ansible_loop_var": "item",
    "changed": false,
    "item": "sshd_service"
}
```
This is because https://github.com/willshersystems/ansible-sshd/pull/199/files#diff-3ca6b148feb962b6f4e0c2a016f5f5f22d3b881aa387a29ef466be43fdd537f5R2
```
- name: Always reset overridden OS defaults
  ansible.builtin.set_fact:
    sshd_clear_role_vars: true
```

Jakuje:
> I cannot fix the failing ansible-lint test. The master branch has the same error, so I guess this PR is ready to review.

These are fixed by #198 already (known ansible(-lint) bug if I see right).

defaults/main.yml
```diff
@@ -73,18 +57,27 @@ sshd_hostkey_mode: "{{ __sshd_hostkey_mode }}"
...
+# more than once. This is fail-safe that will force you to redefine the
+# following variables if you need to set them to a value different to the OS
+# default. See https://github.com/ansible/proposals/issues/171 for more
+# information. This also means that if you use the `roles:` playbook keyword to
+# run the role multiple times, you will need to define your vars at the role
+# level and not in a `vars` playbook keyword, e.g.
+# ```
+# - hosts: all
+#   roles:
+#     - role: ansible-role-ssh
```
Jakuje, line 70:
```suggestion
#     - role: ansible-sshd
```

defaults/main.yml
```diff
@@ -73,18 +57,27 @@ sshd_hostkey_mode: "{{ __sshd_hostkey_mode }}"
...
+#       sshd_config_file: ...
+# `
```
Jakuje, line 72:
```suggestion
# ```
```

meta/10_top.j2
```diff
@@ -20,7 +20,7 @@
 {%     set value = override %}
 {%   elif sshd[key] is defined %}
 {%     set value = sshd[key] %}
-{%   elif __sshd_drop_in_dir and sshd_config_file.startswith(__sshd_drop_in_dir)  %}
+{%   elif __sshd_drop_in_dir and sshd_config_file.startswith(__sshd_drop_in_dir) %}
```
Jakuje, line 23: this does not look like a change related to the original change. Can you separate these at least into separate commit?

Jakuje (commented):
Can you provide a some examples of a playbook that did not work for you, ideally as part of test case here, which could be used as a regression test to make sure this keeps working in the future?

nkakouros:
> This breaks some integration tests - the handler that starts/restarts sshd fails because `sshd_service` is undefined.

How can I run these tests?

richm:
> > This breaks some integration tests - the handler that starts/restarts sshd fails because `sshd_service` is undefined.
> 
> How can I run these tests?

I'm not sure what the ansible-sshd authors would recommend.  The linux system roles team is based on Fedora as the development platform.  We use tox and tox-lsr to run the tests using qemu/kvm.  https://linux-system-roles.github.io/contribute.html
You'll also need a tox.ini like this since ansible-sshd doesn't have one:
```
[lsr_config]
lsr_enable = true

[lsr_ansible-lint]
configfile = {toxinidir}/.ansible-lint
```
With this, I can run the tests like this:
```
tox -e qemu-ansible-core-2.13 -- --image-name centos-9 tests/tests_precedence.yml
```
To run the test playbook `tests/tests_precedence.yml` using ansible-core 2.13 against a centos-9 local VM image.

nkakouros:
I made some changes to address the error with the undefined name. Now, the unsetting of variables is the last thing that happens in the role. It also happens only when it is absolutely needed. I will try to come up with some test case.

meta/10_top.j2
nkakouros: Done.

richm:
I still had some failures with https://github.com/willshersystems/ansible-sshd/pull/199/commits/438e010f2d12f5374f0618d468e07edb4de88a41 so I tried to fix it: https://github.com/richm/ansible-sshd/tree/revert-richm
With this code, all of the tests pass.
One thing I don't understand - why does the role take such great pains to reset the public API variables?  e.g. https://github.com/richm/ansible-sshd/tree/revert-richm - what is the use case for this?  Something like
```yaml
include_role:
  name: sshd
vars:
  sshd_config_file: /path/to/1
...
include_role:
  name: sshd
vars:
  sshd_service: my_custom_sshd_service
```
Where the value `sshd_config_file: /path/to/1` will still be in effect for the second run of the role?

vars/FreeBSD.yml
```diff
@@ -1,7 +1,7 @@
 ---
 __sshd_config_group: wheel
 __sshd_config_mode: "0644"
-sshd_sftp_server: /usr/libexec/sftp-server
 __sshd_defaults:
-  Subsystem: "sftp {{ sshd_sftp_server }}"
+  Subsystem: "sftp {{ __sshd_sftp_server }}"
+__sshd_sftp_server: /usr/libexec/sftp-server
```
Jakuje, line 6: i am not sure if the order here matters, but I do not have a way to test it on freebsd.

defaults/main.yml
Jakuje, line 72:
```suggestion
#     - role: ansible-sshd
#       sshd_config_file: ...
# ```
```
This got broken again

vars/FreeBSD.yml
nkakouros: It shouldn't matter, but I reverted the order to avoid confusion.

nkakouros:
The reason I reset the public API variables is because in my implementation they become facts. According to [Ansible's variable precedence rules](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#ansible-variable-precedence), facts take precedence over playbook vars and others. This means that sth like the below snippet will fail:

``` yaml
- hosts: all
  vars:
    sshd_config_file: /custom/config1
  roles:
    - ansible-sshd

- hosts: all
  roles:
    - ansible-sshd 
    # user expects to run with OS default for sshd_config_file,
    # but /custom/config1 is used instead. Here is where resetting
    # the variables to None is useful.

- hosts: all
  vars:
    sshd_config_file: /custom/config2  # this has no effect
  roles:
    - ansible-sshd
```

> I still had some failures with [438e010](https://github.com/willshersystems/ansible-sshd/commit/438e010f2d12f5374f0618d468e07edb4de88a41) so I tried to fix it: [https://github.com/richm/ansible-sshd/tree/revert-richm](https://github.com/richm/ansible-sshd/tree/revert-richm?rgh-link-date=2022-08-16T23%3A51%3A04Z) 

You are right, using intermediate variables or boolean expressions is the proper solution as it would cover the third case above as well and not need unsetting the variables at all. I started with the assumption that using intermediate variables or logic expression would make the code hard to read and maintain, but not using them breaks user expectations. I will rewrite the PR using intermediate variables.

nkakouros:
I completely rewrote the PR. The issue was actually much simpler, some defaults (e.g. `sshd_service`) were not set correctly, making it impossible for the user to override them. I fixes those and I also add it a test to check for multiple runs of the role. I also moved internal variables out of `defaults/main.yml`.

nkakouros:
BTW, should `__sshd_drop_in_dir` and `__sshd_main_config_file` also be overrideable by the user so that they can create drop in files as they need?

defaults/main.yml
```diff
@@ -47,44 +46,31 @@ sshd_config_file: "{{ __sshd_config_file }}"
 
 ### VARS DEFAULTS
 ### The following are defaults for OS specific configuration in var files in
-### this role. They should not be set directly by role users.
-sshd_packages: []
+### this role. They should not be set directly by role users, unless they know
+### what they are doing, e.g. installing extra packages or installing an
+### alternative SystemD service file
+sshd_packages: "{{ __sshd_packages }}"
```
richm, line 52: Yeah, I forgot about how variables work - they are evaluated lazily - so as long as you don't try to use `sshd_packages` before `__sshd_packages` is defined, it will work - and, it allows users to override `sshd_packages` - nice!

tests/tasks/restore.yml
```diff
@@ -26,10 +26,8 @@
     - __sshd_test_backup.path is defined
 
 - name: Include common variables
-  ansible.builtin.include_role:
-    name: ansible-sshd
-    tasks_from: set_common_vars.yml
-    public: true
+  ansible.builtin.include_vars:
+    file: ../../vars/main.yml
```
richm, line 30: This will break when the role is converted to collection format - that's why we used `include_role` with `tasks_from` because it will work in both the legacy role format and the collection format

When the role is converted to collection format the layout will look like this:
```
namespace/collection_name/
  roles/
    sshd/
      vars/
        main.yml
  tests/
    sshd/
      tasks/
        restore.yml
```
so doing ../../vars/main.yml from here will result in error `ENOFILE`

In addition - when we (system roles team) test the collection format, we move the tests directory completely outside of the collection directory - if you run the tests from inside the collection directory, some implicit path look ups will work that will not work if the playbook is run from outside of the collection directory - so in order to ensure we are testing as if a user is running a playbook from a completely different location, we move the collection `tests/` to e.g. `/tmp` and run the test playbooks from there.  In that case, there is really no easy way to reconstruct the path to the role `vars/main.yml` - but `include_role` with `tasks_from` will still work.

tests/tests_duplicate_role.yml
```diff
@@ -0,0 +1,61 @@
...
+  tasks:
+    - name: Get config1 stat
+      ansible.builtin.stat:
+        path: /etc/ssh/dup_config1
+      register: config1_stat
+
+    - name: Check config1 exists
+      ansible.builtin.assert:
+        that:
+          - "config1_stat.stat.exists"
```
richm, line 38: This doesn't need to be quoted 1) yaml will parse this as a string implicitly 2) if ansible can determine that the context is a jinja statement or expression, it doesn't need to be quoted (unless you are doing something like `- "Some string" in result.output` in which case since the first character is a quoted string, the entire thing has to be quoted like `- '"Some string" in result.output'` or 
```yaml
 - >-
   "Some string" in result.output
```

tests/tasks/restore.yml
nkakouros: I reverted to include_role.

tests/tests_duplicate_role.yml
nkakouros: I removed the quotes.

tests/tasks/restore.yml
nkakouros: A question about errors I encountered while dealing with this. Do you indeed need to support CentOS 6 and Ansible <2.7? All these are end of life.

richm: EL6 - yes - note that even though CentOS 6 may be EOL, RHEL6 is still supported for extended support customers

Ansible < 2.9 - no - only ansible 2.9 and later

nkakouros: Centos6 (the docker image for the tests at least) runs ansible 2.6, so this makes it an implicit requirement to support that version as well?

richm: > Centos6 (the docker image for the tests at least) runs ansible 2.6, so this makes it an implicit requirement to support that version as well?

Hmm - not sure how these CI tests work.  The sshd role should support RHEL6 **managed** hosts, definitely not **controller** hosts (the host where `ansible-playbook` is running).  The sshd role should support running ansible >= 2.9 on controller hosts that are able to run ansible 2.9 (e.g. EL7 can run ansible 2.9, but EL7 cannot run ansible-core 2.12 or later).

richm: > > Centos6 (the docker image for the tests at least) runs ansible 2.6, so this makes it an implicit requirement to support that version as well?
> 
> Hmm - not sure how these CI tests work. The sshd role should support RHEL6 **managed** hosts, definitely not **controller** hosts (the host where `ansible-playbook` is running).

I should qualify that by saying that it is possible there is some user out there somewhere who
* is unwilling or unable to upgrade the EL6 controller host to EL7 or later
* is quite happily using the unsupported, EOL ansible 2.6
* will be impossible to notify of the deprecation of EL6 controller node support
* will be quite unhappy when they update their ansible-sshd role to discover it is now broken

>The sshd role should support running ansible >= 2.9 on controller hosts that are able to run ansible 2.9 (e.g. EL7 can run ansible 2.9, but EL7 cannot run ansible-core 2.12 or later).

nkakouros: Indeed, the tests use the single docker image per OS to run the tests against the localhost. So, the same machine is both the controller and the managed host.

nkakouros: Since Centos 6 uses Ansible 2.6, `include_role` does not support the `public: true` parameter. This for some strange reason does not emit an error in the `tests/tasks/restore.yml` file that uses that parameter but does emit an error if this parameter is used elsewhere. For this reason, I selected to expose as facts some internal variables that the tests need access to. This is done conditionally, if the role runs as part of the test suite.

richm:
lgtm - works locally using qemu/kvm tests - other than the ansible-lint issues, most of which are fixed by https://github.com/willshersystems/ansible-sshd/pull/198, I think this is good to go

Jakuje:
> BTW, should `__sshd_drop_in_dir` and `__sshd_main_config_file` also be overrideable by the user so that they can create drop in files as they need?

Probably yes, but I think we did not get a request for that yet.

tasks/variables.yml
```diff
@@ -24,7 +27,16 @@
         - "{{ ansible_os_family }}_{{ ansible_distribution_major_version }}.yml"
         - "{{ ansible_distribution }}.yml"
         - "{{ ansible_os_family }}.yml"
-        - default.yml
+        - main.yml  # fallback, vars/main.yml is always loaded by Ansible
       paths:
         - "{{ role_path }}/vars"
         - "{{ playbook_dir }}/vars"
+
+- name: Convert variables to facts for testing with old Ansible versions (pre 2.7)
```
Jakuje, line 35: The minimum supported ansible version should be 2.9:
https://github.com/willshersystems/ansible-sshd/blob/master/meta/main.yml#L8

nkakouros: Right. But Centos 6 comes with Ansible 2.6 which makes the tests fail, unless I have misunderstood how the tests are setup.

Jakuje: CentOS 6 is officially supported only as a target host, not as a controlling host. For CI we provide several workarounds already so I would either move this to these workaround or do some condition to run this only when the ansible version is smaller than 2.7.
https://github.com/willshersystems/ansible-sshd/blob/b75ef4c630d37f2f0be0e38cca27bb9518616c4e/.github/workflows/ansible-centos-check.yml#L12

nkakouros: OK, I will add some workaround.

nkakouros: A question, I can't figure out how the following task succeeds when CentOS 6 is used:

https://github.com/willshersystems/ansible-sshd/blob/b75ef4c630d37f2f0be0e38cca27bb9518616c4e/tests/tasks/restore.yml#L34-L43

How is `__sshd_skip_virt_env` defined?

Jakuje: It was defined in the https://github.com/willshersystems/ansible-sshd/blob/master/vars/common.yml, which should be loaded

richm: not sure - the `public: true` option was added in 2.7 - maybe in 2.6 all of the role variables were public by default?  and `public: true` was added in 2.7 so that you could make the role variables private?

nkakouros: Apparently vars included with `include_vars` in the role are actually facts, thus they follow the host and are available to tasks after `include_role`. So, `public: true` is not really needed in the current implementation in `master`. I will come up with a workaround for CentOS 6

.github/workflows/ansible-centos-check.yml
```diff
@@ -10,7 +10,14 @@ jobs:
...
     - run: "sed -i -e 's/meta: end_host/assert:\\n    that: __sshd_os_supported|bool/' tasks/install.yml"
-    - run: "sed -i -e 's/.*public: true//' tests/tasks/restore.yml"
+    - run: >-
+        printf '%s\n%s\n%s\n%s'
+        '- name: Convert variables to facts for testing with old Ansible version (pre 2.7)'
+        '  ansible.builtin.set_fact:'
+        '    __sshd_skip_virt_env: "{{ __sshd_skip_virt_env }}"'
+        '    __sshd_config_file: "{{ __sshd_config_file }}"'
+        >> tasks/variables.yml
+    - run: "sed -i -e 's/.*public: true//' tests/tasks/restore.yml tests/tests_duplicate_role.yml"
```
richm, line 20: it would be better to use `-e '/public: true/d'` - that is, delete the line containing "public: true" - that way if someone changes it so that `public: true` is not the last parameter like
```
- name: Include common variables
  ansible.builtin.include_role:
    name: ansible-sshd
    public: true
    tasks_from: set_common_vars.yml
```
it won't break

nkakouros: Done.

richm (approved):

richm:
tests pass locally with qemu + ansible-core 2.13 + centos-9 vm

Jakuje (commented):
hopefully last comment regarding the test backup. The rest looks good.

tests/tests_duplicate_role.yml
```diff
@@ -0,0 +1,65 @@
+---
+- hosts: all
+  vars:
+    sshd_config_file: /etc/ssh/dup_config1
+  tasks:
+    - name: "Backup configuration files"
+      ansible.builtin.include_tasks: tasks/backup.yml
```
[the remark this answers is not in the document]
nkakouros: I think I added the correct files.

Jakuje (approved):

Jakuje:
I think we are good here. @mattwillsher what do you think? Ready to merge?

mattwillsher:
I'm happy if you all are - I can follow the reasoning for the change. 

Please can someone produce a changelog for this please and I'll get it up to Galaxy. 


Off topic but I've been considering a rewrite of the module to try and consolidate they many changes that have occurred to this code base over the years. I'm getting a little concerned about general stability and side effects of changes.

Jakuje:
> Please can someone produce a changelog for this please and I'll get it up to Galaxy.

The only change since last release was the #201 fixing the auto-tag so I would wait for at least #200 to finalize before doing a new release. For a oneliner, this basically fixes #170 and similar cases, so something like

> * Allow overriding less common role variables `sshd_packages`, `sshd_binary`, `sshd_service` and `sshd_sftp_server` by role user.

> Slightly off-topic but I've been considering a rewrite of the module to try and consolidate the many changes that have occurred to this code base over the years. I'm getting a little concerned about general stability and side effects of changes.

I am not sure if a module would help here much, but I dont have much practical experience with modules. For the stability and side effects of the changes, we have tests coverage, but indeed there might slip some issues such as this one.

mattwillsher:
I mean role (too much Terraform for me these days!).

Perhaps things are ok, I just need to spend more time with the code

PR #200, opened 2022-08-17
Makes drop-in functionality configurable by the user

nkakouros:
This is based off of #199.

This PR simplifies the logic behind the drop-in config files and also allows the user to use drop-in configs even if the distribution does not support it out of the box.

### DIFF
--- a/tasks/install_config.yml
+++ b/tasks/install_config.yml
@@ -1,4 +1,13 @@
 ---
+- name: Create a directory for drop-in configuration snippets
+  ansible.builtin.file:
+    path: "{{ sshd_config_file | dirname }}"
+    state: directory
+    mode: "{{ sshd_drop_in_dir_mode }}"
+  when:
+    - sshd_main_config_file is not none
+    - sshd_config_file | dirname != sshd_main_config_file | dirname
+
 - name: Create the complete configuration file
   ansible.builtin.template:
     src: sshd_config.j2
@@ -18,8 +27,8 @@
 - name: Make sure the include path is present in the main sshd_config
   ansible.builtin.lineinfile:
     insertbefore: BOF
-    line: "Include {{ __sshd_defaults['Include'] }}"
-    path: "{{ __sshd_main_config_file }}"
+    line: "Include {{ sshd_config_file | dirname }}/*.conf"
+    path: "{{ sshd_main_config_file }}"
     owner: "{{ sshd_config_owner }}"
     group: "{{ sshd_config_group }}"
     mode: "{{ sshd_config_mode }}"
@@ -32,7 +41,5 @@
     backup: "{{ sshd_backup }}"
   notify: reload_sshd
   when:
-    - __sshd_defaults['Include'] | d(false)
-    - __sshd_main_config_file is not none
-    - __sshd_drop_in_dir is not none
-    - sshd_config_file.startswith(__sshd_drop_in_dir)
+    - sshd_main_config_file is not none
+    - sshd_config_file | dirname != sshd_main_config_file | dirname
--- a/templates/sshd_config.j2
+++ b/templates/sshd_config.j2
@@ -21,8 +21,9 @@
 {%     set value = override %}
 {%   elif sshd[key] is defined %}
 {%     set value = sshd[key] %}
-{%   elif __sshd_drop_in_dir and sshd_config_file.startswith(__sshd_drop_in_dir)  %}
-{#     The drop-in directory does not use the defaults from main file to avoid recursion #}
+{%   elif sshd_main_config_file is not none
+       and sshd_config_file | dirname != sshd_main_config_file | dirname %}
+{#     Do not use the defaults from main file to avoid recursion #}
 {%   elif __sshd_defaults[key] is defined and not sshd_skip_defaults %}
 {%     if key == 'HostKey' and __sshd_fips_mode %}
 {%       set value = __sshd_defaults[key] | difference(__sshd_hostkeys_nofips) %}
--- a/templates/sshd_config_snippet.j2
+++ b/templates/sshd_config_snippet.j2
@@ -20,8 +20,9 @@
 {%     set value = override %}
 {%   elif sshd[key] is defined %}
 {%     set value = sshd[key] %}
-{%   elif __sshd_drop_in_dir and sshd_config_file.startswith(__sshd_drop_in_dir)  %}
-{#     The drop-in directory does not use the defaults from main file to avoid recursion #}
+{%   elif sshd_main_config_file is not none
+       and sshd_config_file | dirname != sshd_main_config_file | dirname %}
+{#     Do not use the defaults from main file to avoid recursion #}
 {%   elif __sshd_defaults[key] is defined and not sshd_skip_defaults %}
 {%     if key == 'HostKey' and __sshd_fips_mode %}
 {%       set value = __sshd_defaults[key] | difference(__sshd_hostkeys_nofips) %}
--- a/tests/tests_include_present.yml
+++ b/tests/tests_include_present.yml
@@ -81,3 +81,85 @@
 
     - name: "Restore configuration files"
       ansible.builtin.include_tasks: tasks/restore.yml
+
+- hosts: all
+  vars:
+    __sshd_test_backup_files:
+      - /etc/ssh/custom_sshd_config
+      - /etc/ssh/custom_sshd_config.d/custom-drop-in
+
+  tasks:
+    - name: Run only for a specific OS with an capable OpenSSH version
+      ansible.builtin.meta: end_play
+      when:
+        ansible_facts['distribution'] != 'Ubuntu'
+        or ansible_facts['distribution_major_version']|int != 20
+
+    - name: "Backup configuration files"
+      ansible.builtin.include_tasks: tasks/backup.yml
+
+    - name: Create dummy main configuration file
+      # Normally, this should not be needed. For test, however, we need a file
+      # different to the one in the first play.
+      file:
+        path: /etc/ssh/custom_sshd_config
+        state: touch
+
+    - name: Create a new configuration in a custom drop-in directory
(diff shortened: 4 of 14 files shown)


mattwillsher:
@nkakouros lint should be fixed on master I believe, so please merge/rebash that in to the test work.

This change supercedes #199 ?

nkakouros:
@mattwillsher Sorry, I was too sleepless to read properly, this is supposed to be a continuation of #199. I updated the title. I will rebase both PRs now.

templates/sshd_config_snippet.j2
```diff
@@ -6,10 +6,10 @@
 {%     elif value is sameas false %}
 {{ key }} no
 {%     elif value is string or value is number %}
-{{ key }} {{ value | string }}
+{{ key }} {{ value }}
```
richm, line 9: Why is this being reverted?

templates/sshd_config_snippet.j2
```diff
@@ -6,10 +6,10 @@
...
 {%     else %}
 {%       for i in value %}
-{{ key }} {{ i | string }}
+{{ key }} {{ i }}
```
richm, line 12: Why is this being reverted?

tests/tasks/restore.yml
```diff
@@ -28,8 +28,7 @@
 - name: Include common variables
   ansible.builtin.include_role:
     name: ansible-sshd
-    tasks_from: set_common_vars.yml
-    public: true
+    tasks_from: variables.yml
```
richm, line 31: Removing `public: true` will likely break tests that need access to internal role variables for verification.

Instead - we may need two tasks here:
```
- name: Include common variables
  ansible.builtin.include_role:
    name: ansible-sshd
    public: true
    tasks_from: variables.yml
  when: ansible_version.full is version('2.6', '>')

- name: Include common variables
  ansible.builtin.include_role:
    name: ansible-sshd
    tasks_from: variables.yml
  when: ansible_version.full is version('2.7', '<')
```

templates/sshd_config.j2
```diff
@@ -7,10 +7,10 @@
 {%     elif value is sameas false %}
 {{ key }} no
 {%     elif value is string or value is number %}
-{{ key }} {{ value | string }}
+{{ key }} {{ value }}
```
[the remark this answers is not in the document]
nkakouros: I have no idea 8-|. I fixed that.

tasks/install_config.yml
```diff
@@ -18,8 +27,8 @@
 - name: Make sure the include path is present in the main sshd_config
   ansible.builtin.lineinfile:
     insertbefore: BOF
-    line: "Include {{ __sshd_defaults['Include'] }}"
-    path: "{{ __sshd_main_config_file }}"
+    line: "Include {{ sshd_config_file | dirname }}/*.conf"
```
Jakuje, line 30: I am a bit worried that in the future some distribution will come up with different extension or drop-in directory, which will cause issues of hardcoding the `*.conf` here.  Its good that this did not happen yet with both Fedora and Ubuntu supporting the same path and extension.

tasks/install_config.yml
```diff
@@ -1,4 +1,13 @@
 ---
+- name: Create config directory in directory
```
Jakuje, line 2: Maybe a better wording, something along the lines like to be more clear what does this task do?

> Create a directory for drop-in configuration snippets

tasks/install_config.yml
nkakouros: Do you suggest setting to just `*` instead of `*.conf`?

tasks/install_config.yml
nkakouros: Done.

tasks/install_config.yml
nkakouros: I also thought of doing:

```yaml
- name: Make sure the include path is present in the main sshd_config          
  ansible.builtin.lineinfile:                                                  
    insertbefore: BOF                                                          
    line: "Include {{ include_path }}"                                  
    path: "{{ sshd_main_config_file }}"                                        
    owner: "{{ sshd_config_owner }}"                                          
    group: "{{ sshd_config_group }}"                                           
    mode: "{{ sshd_config_mode }}"                                            
    validate: >-                                                               
      {% if sshd_test_hostkey is defined and sshd_test_hostkey.path is defined %}
        {{ sshd_binary }} -t -f %s -h {{ sshd_test_hostkey.path }}/rsa_key     
      {% else %}                                                              
        {{ sshd_binary }} -t -f %s                                             
      {% endif %}                                                              
    backup: "{{ sshd_backup }}"                                                
  notify: reload_sshd                                                          
  vars:                                                                        
    include_path: "{{ __sshd_defaults.Include | d(sshd_config_file | dirname + '/*') }}"
  when:                                                                      
    - sshd_main_config_file is not none                                        
    - sshd_config_file | dirname != sshd_main_config_file | dirname            
```

But I don't think it makes much sense. In fact, I was considering removing the `Include` directives from `__sshd_defaults` but I decided not to, so that the role will continue replicating the OS defaults exactly by default.

Jakuje: > Do you suggest setting to just `*` instead of `*.conf`?

No, the `*.conf` is better than the `/*` as it reflects the current configuration in distributions. The only problem will be when somebody will want to have it configurable, either on some different distro or in their playbook.

The drop-in directory is implemented in OpenSSH as a generic include directive, where the glob does not necessarily need to be just in the filename path, but it can be anything like `/etc/ssh/sshd_config.d/*/sshd_config` when you would like to organize your snippets in separate directories, which would break your assumptions, in both of your suggested changes, which I believe is not necessary and which would advocate keeping the above line as it was

tests/tests_include_present.yml
```diff
@@ -81,3 +81,73 @@
...
+            that:
+              - "'Banner /etc/include-issue' not in custom_config.content | b64decode"
+              - "'Ciphers aes192-ctr' not in custom_config.content | b64decode"
+              - "'Include /etc/ssh/custom_sshd_config.d/*.conf' in custom_config.content | b64decode"
+
+      tags: tests::verify
+
+    - name: "Restore configuration files"
+      ansible.builtin.include_tasks: tasks/restore.yml
+
```
Jakuje, line 153: can you remove the trailing newline to fix the ansible lint check

tests/tests_duplicate_role.yml
```diff
@@ -0,0 +1,65 @@
+---
+- hosts: all
+  vars:
+    sshd_config_file: /etc/ssh/dup_config1
+  tasks:
+    - name: "Backup configuration files"
+      ansible.builtin.include_tasks: tasks/backup.yml
```
Jakuje, line 7: I thought I was posting this before, but probably in the other PR. The backup tasks do not do anything unless `__sshd_test_backup_files` is defined, which I do not see anywhere in this file. Please, list in this list any file you might modify in this test.

nkakouros: I rebased on the other PR's branch, it should be fixed now.

tests/tests_include_present.yml
nkakouros: Done.

tasks/install_config.yml
nkakouros: If a user uses the role to manage 100% of their sshd setup, I think we don't have an issue. If the user has a lot of drop in config files under different directories, they will need to run the role multiple times, each time replacing the `sshd_config_file` variable. In all of these cases, a new `Include` directive will be added on top of any previous ones. If the need a custom `Include` directive (e.g. `/etc/ssh/sshd_config.d/*/sshd_config` from your example), they can first run the role without a drop-in config to make use of the `sshd_config` template that will allow them to set an arbitrary `Include`, and then run the role as many times as the drop in configs they have (or, better, use a separate role/task to upload the drop in files en masse).

tests/tasks/restore.yml
nkakouros: This is resolved (by rebasing on #199 ).

tasks/install_config.yml
nkakouros: Done.

vars/main.yml
```diff
@@ -31,15 +31,11 @@ __sshd_os_supported: no
...
 __sshd_sysconfig_supports_use_strong_rng: false
 
-__sshd_runtime_directory: false
+__sshd_runtime_directory: ~
 __sshd_runtime_directory_mode: "0755"
 
-# If the system supports drop-in directory, it is configured in this variable. It is used
-# to distinguish if we are writing a configuration snippet or we should write defaults.
-__sshd_drop_in_dir: false
-# this is the path to the main sshd_config which is checked for Include directive when
```
Jakuje, line 40: I think the last line with comment should not have been removed or the following should have been removed too.

tasks/install_config.yml
```diff
@@ -1,4 +1,13 @@
 ---
+- name: Create a directory for drop-in configuration snippets
+  ansible.builtin.file:
+    path: "{{ sshd_config_file | dirname }}"
+    state: directory
+    mode: "0755"
```
Jakuje, line 6: one more thing, the permissions for the directory, at least in fedora, is 0700. We should keep this permission or make it configuratable (with sane defaults). Additionally some assert checking the permissions before and after might make sense too, for example in the `tests_os_defaults.yml`.

mattwillsher: If this varies per OS the default should come from the OS specific file and maybe have an override, particularly if some variants have more open than 0700.

vars/main.yml
nkakouros: Done.

tasks/install_config.yml
nkakouros: Done.

Jakuje (approved):

Jakuje:
Thank you for the patience! It looks good now.

PR #201, opened 2022-08-18
parse [vX.Y.Z] correctly from CHANGELOG.md

richm:


richm:
Fixes https://github.com/willshersystems/ansible-sshd/pull/195#issuecomment-1219479887

.github/workflows/changelog_to_tag.yml
```diff
@@ -56,3 +74,8 @@ jobs:
           body_path: ./.tagmsg.txt
           draft: false
           prerelease: false
+      - name: Publish role to Galaxy
+        uses: robertdebock/galaxy-action@1.2.0
+        with:
+          galaxy_api_key: ${{ secrets.galaxy_api_key }}
+          git_branch: ${{ steps.tag.outputs.branch }}
```
Jakuje, line 81: How is the galaxy publishing going to work here. This requires the secrets to be set up, which I think is not the case right now. Otherwise it looks ok.

Jakuje (approved):

.github/workflows/changelog_to_tag.yml
richm: ah, right - I'll remove this

PR #204, opened 2022-09-07
keep v prefix in version/tag

richm:
Keep the `v` prefix in the version/tag


richm:
This should fix the bug in https://github.com/willshersystems/ansible-sshd/runs/8222743784?check_suite_focus=true

PR #205, opened 2022-09-14
Adding support for OpenWrt

sscheib:
With this PR I'd like to provide support for OpenWrt. It has been tested on OpenWrt 22.03.

Please let me know if anything is missing.

### DIFF
--- a/vars/OpenWrt.yml
+++ b/vars/OpenWrt.yml
@@ -1,0 +1,14 @@
+---
+__sshd_service: sshd
+__sshd_packages:
+  - openssh-server
+  - openssh-sftp-server
+__sshd_config_mode: "0644"
+__sshd_sftp_server: "/usr/lib/sftp-server"
+__sshd_defaults:
+  Subsystem: "sftp {{ __sshd_sftp_server }}"
+  AuthorizedKeysFile: .ssh/authorized_keys
+  HostKey:
+    - /etc/ssh/ssh_host_rsa_key
+    - /etc/ssh/ssh_host_ed25519_key
+__sshd_os_supported: yes
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Tested on:
 * FreeBSD 10.1
 * OpenBSD 6.0
 * AIX 7.1, 7.2
+* OpenWrt 21.03
 
 It will likely work on other flavours and more direct support via suitable
 [vars/](vars/) files is welcome.
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -9,6 +9,7 @@
     - ansible_virtualization_type|default(None) not in __sshd_skip_virt_env
     - ansible_connection != 'chroot'
     - ansible_os_family != 'AIX'
+    - ansible_os_family != 'OpenWrt'
   listen: reload_sshd
 
 # sshd on AIX cannot be 'reloaded', it must be Stopped+Started.
@@ -29,3 +30,13 @@
   when:
     - sshd_allow_reload|bool
     - ansible_os_family == 'AIX'
+
+# sshd on OpenWrt does not support reloading a service, it has to be restarted instead
+- name: Reload the SSH service (OpenWrt)
+  ansible.builtin.service:
+    name: "{{ sshd_service }}"
+    state: restarted
+  when:
+    - sshd_allow_reload|bool
+    - ansible_os_family == 'OpenWrt'
+  listen: reload_sshd


vars/OpenWrt.yml
```diff
@@ -0,0 +1,11 @@
...
+__sshd_service: sshd
+__sshd_packages:
+  - openssh-server
+  - openssh-sftp-server
+__sshd_config_mode: "0644"
+__sshd_defaults:
+  Subsystem: "sftp {{ __sshd_sftp_server }}"
+  AuthorizedKeysFile: .ssh/authorized_keys
+__sshd_os_supported: yes
+__sshd_runtime_directory: /var/run/
```
Jakuje, line 11: Does the openWRT need a runtime directory? I do not see this path mentioned anywhere in the https://github.com/openwrt/packages/tree/openwrt-21.02/net/openssh

vars/OpenWrt.yml
```diff
@@ -0,0 +1,11 @@
+---
+__sshd_service: sshd
+__sshd_packages:
+  - openssh-server
+  - openssh-sftp-server
+__sshd_config_mode: "0644"
+__sshd_defaults:
+  Subsystem: "sftp {{ __sshd_sftp_server }}"
```
Jakuje, line 8: Are there any other defaults in openwrt?

vars/OpenWrt.yml
sscheib: Right, that is probably a misunderstanding on my end. I thought of it as "the place where the PID goes" and indeed while keeping everything as default, sshd on OpenWrt will create a PID in /var/run/sshd.pid.
Generally speaking, every runtime-related data should be placed in /var/run, as devices which are running OpenWrt typically are embedded devices with limited read/write cycles on the EPROM. /var/run is mounted as tmpfs and thus no read/write cycles are performed on the EPROM.
If I have a misunderstanding regarding that runtime directory, please let me know. I was just trying to be sensitive in terms of runtime data (due the limitations in read/write cycles).

Jakuje: Right. It is probably not well documented.

But this option is used by the `systemd` as a working directory on Ubuntu-based systems (its missing from the service template though):

https://github.com/willshersystems/ansible-sshd/blob/master/vars/main.yml#L34
https://github.com/willshersystems/ansible-sshd/blob/master/vars/Ubuntu_22.yml#L20
https://github.com/willshersystems/ansible-sshd/blob/master/templates/sshd.service.j2#L13
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=

I think you should be good by excluding this option.

vars/OpenWrt.yml
sscheib: I simply installed openssh-server on a freshly installed on OpenWrt 21.02 and removed all things which were commented (thus, I assume, are set by default without explicitly stating them in the sshd_config). What remains is the following:
`root@OpenWrt:~# grep -Ev '#|^$' /etc/ssh/sshd_config
HostKey /etc/ssh/ssh_host_rsa_key
HostKey /etc/ssh/ssh_host_ed25519_key
AuthorizedKeysFile      .ssh/authorized_keys
Subsystem       sftp    /usr/lib/sftp-server`

However, I assume the HostKey entries are also baked into the "default" configuration as removing them from the configuration file will work without any issue.
If there is a better way of determining what's a default, please let me know.

vars/OpenWrt.yml
sscheib: Thanks for letting me know - I corrected it :).

vars/OpenWrt.yml
Jakuje: That is a good way to get the defaults. Please, include also the hostkeys, so they can be checked for existence before checking the configuration is valid by the role.
I also see that the path to the sftp is not default so please, provide also a definition of  `__sshd_sftp_server`.

sscheib: Both fixed, thanks! :)

sscheib:
I added another handler (same approach as with AIX), as sshd does not implement reloading of the configuration file.
Down below a quick demonstration:
```
# currently running on port 1905
root@OpenWrt:~# netstat -tulpen | grep 1905
tcp        0      0 0.0.0.0:1905            0.0.0.0:*               LISTEN      22831/sshd -D [list

# changing port to 1906
root@OpenWrt:~# grep 1906 /etc/ssh/sshd_config
Port 1906

# "reloading" sshd
root@OpenWrt:~# /etc/init.d/sshd reload

# still listening on 1905 and not on 1906
root@OpenWrt:~# netstat -tulpen | grep 1905
tcp        0      0 0.0.0.0:1905            0.0.0.0:*               LISTEN      22831/sshd -D [list
root@OpenWrt:~# netstat -tulpen | grep 1906

# restarting the sshd
root@OpenWrt:~# /etc/init.d/sshd restart

# now listening on 1906 and not 1905
root@OpenWrt:~# netstat -tulpen | grep 1906
tcp        0      0 0.0.0.0:1906            0.0.0.0:*               LISTEN      22950/sshd -D [list
root@OpenWrt:~# netstat -tulpen | grep 1905
```

While in theory init.d scripts in OpenWrt can choose to implement reload, sshd does not implement it. Only following functions are defined:
- `start_service`
- `shutdown`

A service that supports reloading the configuration is dhcpd (isc-dhcp-server), which implements also `reload_service`.
Unfortunately I did not realize earlier that the reloading did not actually work. OpenWrt silently did nothing :(

Jakuje (approved):
thank you for your contribution! Looks good to me now.

sscheib:
Sorry, whitespaces also fixed, ansible-lint does not report any more errors (on edited/added files).

sscheib:
Last comment, sorry for not adding earlier: Added OpenWrt 21.03 to tested OSes

Jakuje:
> Last comment, sorry for not adding earlier: Added OpenWrt 21.03 to tested OSes

I think we are good now. Last thing, can you squash the fixup commits so I can just merge the stuff without the merge commit?

sscheib:
Should be good now :)

Jakuje:
@mattwillsher any comments or are we good to merge?

mattwillsher:
Ah sorry, I though I'd done this. Look fine to me.

PR #207, opened 2022-09-27
Introduce automatic test for new configuration options in OpenSSH and add missing options so far

### LINKED ISSUE #162: Detect new openssh configuration options more automatically
Adding new options requires reading a diff of openssh sources or manual pages.

I did something similar in the ssh client role, which kind of worked recently:

https://github.com/linux-system-roles/ssh/blob/master/tests/tests_all_options.yml

Or there might be some better way that I did not figure out.

Jakuje:
Fixes #162

### DIFF
--- a/meta/options_body
+++ b/meta/options_body
@@ -28,6 +28,7 @@ DebianBanner
 DenyGroups
 DenyUsers
 DisableForwarding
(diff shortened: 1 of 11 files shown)


mattwillsher (approved):
good improvement, thanks.

PR #211, opened 2022-12-16
Add support for managing selinux and firewall on RHEL

Jakuje:
To simplify supporting running SSHD on different port, we would like to allow the sshd role to manage both SELinux and Firewall on RHEL-based systems (the roles we are using are supported only on these). This is done along the lines with the similar requirements implemented in other roles, for example [cockpit](https://github.com/linux-system-roles/cockpit/pull/73), [postfix](https://github.com/linux-system-roles/postfix/pull/56). It should not have any effect on other OSes.

The tests are currently not executed in the upstream CI, because they are containers without options to customize either firewall or selinux to my best understanding so the test will have to run only downstream with real VMs (unless there is some trick). The provided test is also quite basic, but it can be improved based on the feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,21 @@ Tested on:
 It will likely work on other flavours and more direct support via suitable
 [vars/](vars/) files is welcome.
 
+
+### Optional requirements
+
+If you want to use advanced functionality of this role that can configure
+firewall and selinux for you, which is mostly useful when custom port is used,
+the role requires additional collections which are specified in
+`meta/collection-requirements.yml`. These are not automatically installed.
+You must install them like this:
+```
+ansible-galaxy install -vv -r meta/collection-requirements.yml
+```
+
+For more information, see `sshd_manage_firewall` and `sshd_manage_selinux`
+options below. These roles are supported only on Red Hat based Linux.
+
 Role variables
 ---------------
 
@@ -92,6 +107,25 @@ The templates for the service files to be used are pointed to by the variables
 Using these variables, you can use your own custom templates. With the above
 default templates, the name of the installed ssh service will be provided by
 the `sshd_service` variable.
+
+* `sshd_manage_firewall`
+
+If set to *true*, the the SSH port(s) will be opened in firewall. Note, this
+works only on Red Hat based OS. The default is *false*.
+
+NOTE: `sshd_manage_firewall` is limited to *adding* ports. It cannot be used
+for *removing* ports. If you want to remove ports, you will need to use the
+firewall system role directly.
+
+* `sshd_manage_selinux`
+
+If set to *true*, the the selinux will be configured to allow sshd listening
+on the given SSH port(s). Note, this works only on Red Hat based OS.
+The default is *false*.
+
+NOTE: `sshd_manage_selinux` is limited to *adding* policy. It cannot be used
+for *removing* policy. If you want to remove ports, you will need to use the
+selinux system role directly.
 
 * `sshd`
 
--- a/meta/collection-requirements.yml
+++ b/meta/collection-requirements.yml
@@ -1,4 +1,3 @@
 ---
 collections:
-  - name: ansible.posix
   - name: fedora.linux_system_roles
--- a/tasks/find_ports.yml
+++ b/tasks/find_ports.yml
@@ -1,0 +1,26 @@
+---
+- name: Find the port the ssh service is going to use
+  vars:
+    # This mimics the macro body_option() in sshd_config.j2
+    # The explicit to_json filter is needed for Python 2 compatibility
+    __sshd_ports_from_config_tmp: >-
+      {% if sshd_Port is defined %}
+        {{ sshd_Port | to_json }}
+      {% elif sshd['Port'] is defined %}
+        {{ sshd['Port'] | to_json }}
+      {% elif __sshd_defaults['Port'] is defined and not sshd_skip_defaults %}
+        {{ __sshd_defaults['Port'] | to_json }}
+      {% else %}
+        {{ [22] | to_json }}
+      {% endif %}
+  ansible.builtin.set_fact:
+    __sshd_ports_from_config: >-
+      {% if __sshd_ports_from_config_tmp | from_json is string or __sshd_ports_from_config_tmp | from_json is number %}
+        {{ [__sshd_ports_from_config_tmp | from_json] | to_json }}
+      {% else %}
+        {{ __sshd_ports_from_config_tmp }}
+      {% endif %}
+  when:
+    - sshd_manage_firewall | bool or sshd_manage_selinux | bool
+    - ansible_facts['os_family'] == 'RedHat'
+    - ansible_virtualization_type|default(None) not in __sshd_skip_virt_env
--- a/tasks/firewall.yml
+++ b/tasks/firewall.yml
@@ -1,0 +1,25 @@
+---
+- name: Ensure the ssh service or custom ports are opened in firewall
+  block:
+    - name: Enable the ssh service on default port
+      ansible.builtin.include_role:
+        name: fedora.linux_system_roles.firewall
+      vars:
+        firewall:
+          - service: ssh
+            state: enabled
+      when:
+        - __sshd_ports_from_config | from_json == [22]
+
+    - name: Enable the non-default port(s)
+      ansible.builtin.include_role:
+        name: fedora.linux_system_roles.firewall
+      vars:
+        firewall:
+          - port: "{{ sshd_item }}/tcp"
+            state: enabled
+      loop: "{{ __sshd_ports_from_config | from_json | d([]) }}"
+      loop_control:
+        loop_var: sshd_item  # avoid conflicts with the firewall loops
+      when:
+        - __sshd_ports_from_config | from_json != [22]
--- a/tasks/selinux.yml
+++ b/tasks/selinux.yml
@@ -1,0 +1,16 @@
+---
+- name: Ensure the custom ports are configured in selinux
+  ansible.builtin.include_role:
+    name: fedora.linux_system_roles.selinux
+  vars:
+    selinux_ports:
+      - ports: "{{ sshd_item }}"
+        proto: tcp
+        setype: ssh_port_t
+        state: present
+        local: true
+  loop: "{{ __sshd_ports_from_config | from_json | d([]) }}"
+  loop_control:
+    loop_var: sshd_item  # avoid conflicts with the selinux loops
+  when:
+    - __sshd_ports_from_config | from_json != [22]
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -73,3 +73,11 @@ sshd_hostkey_mode: "{{ __sshd_hostkey_mode }}"
 # instead of replacing the whole configuration file, just add a specified
 # snippet
 sshd_config_namespace: null
+
+# If this option is enabled, the role will configure firewall to open the ports
+# defined in the configuration. This works only on Red Hat based systems.
+sshd_manage_firewall: false
+
+# If this option is enabled, the role will configure selinux to allow sshd to
+# bind the ports defined in the configuration. This works only on Red Hat based systems.
+sshd_manage_selinux: false
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -121,6 +121,24 @@
       when:
         - __sshd_runtime_directory is not none
 
+    - name: Find SSHD ports
+      ansible.builtin.include_tasks: find_ports.yml
+
+    - name: Configure firewall
+      ansible.builtin.include_tasks: firewall.yml
+      when:
+        - sshd_manage_firewall | bool
+        - ansible_facts['os_family'] == 'RedHat'
+        - ansible_facts['distribution_version'] is version('7', '>=')
+        - ansible_virtualization_type|default(None) not in __sshd_skip_virt_env
+
+    - name: Configure selinux
+      ansible.builtin.include_tasks: selinux.yml
+      when:
+        - sshd_manage_selinux | bool
+        - ansible_facts['os_family'] == 'RedHat'
+        - ansible_virtualization_type|default(None) not in __sshd_skip_virt_env
+
     - name: Create the complete configuration file
       ansible.builtin.include_tasks: install_config.yml
       when: sshd_config_namespace is none
--- a/tests/tests_firewall_selinux.yml
+++ b/tests/tests_firewall_selinux.yml
@@ -1,0 +1,127 @@
+---
+- hosts: all
+  vars:
+    __sshd_test_backup_files:
+      - /etc/ssh/sshd_config
+      - /etc/ssh/sshd_config.d/00-ansible_system_role.conf
+  tasks:
+    - name: "Backup configuration files"
+      ansible.builtin.include_tasks: tasks/backup.yml
+
+    ##########
+    # First test: default port
+    ##########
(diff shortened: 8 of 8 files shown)


richm:
@nhosoi ptal

tasks/find_ports.yml
```diff
@@ -0,0 +1,22 @@
...
+    # The explicit to_json filter is needed for Python 2 compatibility
+    __sshd_ports_from_config_tmp: >-
+      {% if sshd_Port is defined %}
+        {{ sshd_Port | to_json }}
+      {% elif sshd['Port'] is defined %}
+        {{ sshd['Port'] | to_json }}
+      {% elif __sshd_defaults['Port'] is defined and not sshd_skip_defaults %}
+        {{ __sshd_defaults['Port'] | to_json }}
+      {% else %}
+        {{ [ "22" ] | to_json }}
```
richm, line 14: I think this is the Jinja spacing it is complaining about - I think it should be `["22"]` - current style guides are to have no spaces between the open and close brackets in lists, and the open and close braces in dicts.

tasks/find_ports.yml
```diff
@@ -0,0 +1,22 @@
...
+      {% endif %}
+  ansible.builtin.set_fact:
+    __sshd_ports_from_config: >-
+      {% if __sshd_ports_from_config_tmp | from_json is string or __sshd_ports_from_config_tmp | from_json is number %}
+        {{ [ __sshd_ports_from_config_tmp | from_json ] | to_json }}
```
richm, line 19:
```suggestion
        {{ [__sshd_ports_from_config_tmp | from_json] | to_json }}
```

README.md
```diff
@@ -27,6 +27,18 @@ via simple password. If you need this functionality, be sure to set
...
+If you want to use advanced functionality of this role that can configure
+firewall and selinux for you, which is mostly useful when custom port is used,
+the role requires additional collections which are specified in
+`meta/collection-requirements.yml`. These are not automatically installed.
+You must install them like this:
+```
+ansible-galaxy install -vv -r meta/collection-requirements.yml
+```
+
+For more information, see `sshd_manage_firewalld` and `sshd_manage_selinux`
```
nhosoi, line 39: ^sshd_manage_firewalld^sshd_manage_firewall^?

README.md
```diff
@@ -93,6 +105,25 @@ Using these variables, you can use your own custom templates. With the above
 default templates, the name of the installed ssh service will be provided by
 the `sshd_service` variable.
 
+* `sshd_manage_firewall`
+
+If set to *true*, the the SSH port(s) will be opened in firewall. Note, this
+works only on Red Hat based OS. The default is *no*.
```
nhosoi, line 111: As stated with `If set to *true*`, we could say `The default is *false*.`?

tasks/find_ports.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Find the port the ssh service is going to use
```
nhosoi, line 2: We may want to skip this finding the port effort unless these conditions are satisfied?
    - sshd_manage_firewall | bool or sshd_manage_selinux | bool
    - ansible_facts['os_family'] == 'RedHat'

README.md
richm: > As stated with `If set to *true*`, we could say `The default is *false*.`?
> 
> Same for `sshd_manage_selinux`?

Yes - please use `true` and `false` instead of `yes` and `no` - this is the current Ansible guideline - https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style#yaml-and-jinja2-syntax "Use `true` and `false` for boolean values in playbooks"

README.md
mattwillsher: Can this be put under a subheading lower down in the requirements? This is not likely to be a requirement for most cases.

README.md
```diff
@@ -93,6 +105,25 @@ Using these variables, you can use your own custom templates. With the above
...
+
+NOTE: `sshd_manage_firewall` is limited to *adding* ports. It cannot be used
+for *removing* ports. If you want to remove ports, you will need to use the
+firewall system role directly.
```
mattwillsher, line 115: What about port 22? That stays open if the port is changed via this role?

README.md
```diff
@@ -93,6 +105,25 @@ Using these variables, you can use your own custom templates. With the above
...
+
+* `sshd_manage_selinux`
+
+If set to *true*, the the selinux will be configured to allow sshd listening
+on the given SSH port(s). Note, this works only on Red Hat based OS.
+The default is *no*.
```
mattwillsher, line 121:
```suggestion
The default is *false*.
```

meta/collection-requirements.yml
```diff
@@ -0,0 +1,3 @@
+---
+collections:
+  - name: fedora.linux_system_roles
```
mattwillsher, line 3: What happens on other distros and OSs it this gets brought in?

tasks/firewall.yml
```diff
@@ -0,0 +1,28 @@
...
+  block:
+    - name: Enable the ssh service on default port
+      ansible.builtin.include_role:
+        name: fedora.linux_system_roles.firewall
+      vars:
+        firewall:
+          - service: ssh
+            state: enabled
+      when:
+        - __sshd_ports_from_config | from_json == [ 22 ]
```
mattwillsher, line 15: This is set as a string via vars above but test as a number?

mattwillsher (commented):
I have some misgivings about OS specific code for other distro components going into this code base directly. It's been a while since I looked, but isn't that what dependenciens via roles handle?

tasks/find_ports.yml
Jakuje: Thanks! Let me try that

README.md
Jakuje: Sure. I think "Optional requirements" subheading would be ok.

Jakuje: > ^sshd_manage_firewalld^sshd_manage_firewall^?

indeed. Thanks for catching that

README.md
Jakuje: Thanks. Not sure how the no ended up here. Probably time for Christmas break.

tasks/find_ports.yml
Jakuje: Good point. Right now it is not used for anything else so we will need all of the three conditions.

README.md
Jakuje: yes. Especially for the SSH port which is usually used for the system management, this would sound quite dangerous. But it could be done if needed.

Jakuje:
The Debian buster action is now broken because of updated git (fix is here https://github.com/roles-ansible/check-ansible-debian-buster-action/pull/2)

Jakuje:
> I have some misgivings about OS specific code for other distro components going into this code base directly. It's been a while since I looked, but isn't that what dependenciens via roles handle?

AFAIK the dependencies via roles can not handle this as the dependencies are being directly executed, which is not what we want here. I see this information missing from https://linux-system-roles.github.io/documentation/role-requirements.html but I know it was mentioned somewhere internally. Would it make sense to mention this also on the page above as a reasoning for the requirements specification @richm ?

richm:
> > I have some misgivings about OS specific code for other distro components going into this code base directly. It's been a while since I looked, but isn't that what dependenciens via roles handle?
> 
> AFAIK the dependencies via roles can not handle this as the dependencies are being directly executed, which is not what we want here.

Right - if you use the standard documented role dependency interface - https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-dependencies - then the roles not only are downloaded, they are executed in the order specified.  For all of the other system roles, *we do not want to do that*, we want to use Ansible constructs like `include_role` to "call" roles only where we want, and conditionally, depending on dynamic conditions using `when`.  We cannot do any of that with the standard role dependency mechanism, which is why we invented our own that @Jakuje linked below.

> I see this information missing from https://linux-system-roles.github.io/documentation/role-requirements.html but I know it was mentioned somewhere internally.

Which information?  The information about the standard Ansible role dependency mechanism?

> Would it make sense to mention this also on the page above as a reasoning for the requirements specification @richm ?

Sure.

meta/collection-requirements.yml
richm: > @richm was trying to describe this in the following page. This file is not processed automatically by the tooling when installing the collection and needs to be installed manually. These two roles do not work anywhere else than on RHEL/Fedora now so for other OSes, this should be just noop if I understand it correctly.
> 
> https://linux-system-roles.github.io/documentation/role-requirements.html

I'm not sure what will happen.  The roles themselves do not have any sort of mechanism to say "Sorry - role not supported on your platform - task skipped".  It is up to whoever uses the role to read the meta/main.yml for the role to determine the platform support, and use `when` conditionals to avoid using the role on unsupported platforms.

For example - the metrics role supports RHEL6 - but the firewall role does not - https://github.com/linux-system-roles/metrics/pull/126/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R114 - when the metrics role uses the firewall role, it has to have a condition to exclude the firewall role on unsupported platforms: https://github.com/linux-system-roles/metrics/blob/main/tasks/firewall.yml#L40-L43
```
  when:
    - metrics_manage_firewall | bool
    - ansible_facts['os_family'] == 'RedHat'
    - ansible_facts['distribution_version'] is version('7', '>=')
```
The sshd role would need a similar conditional for the selinux role (to use only on `ansible_facts['os_family'] == 'RedHat'`) and for the firewall role (probably similar to the above)

README.md
mattwillsher: So sshd will also continue to listen on port 22 for management?
As a user of this module, I would expect the firewall to mirror the port settings provided to the module if changed. I don't see a need to have ssh listen on multiple ports, and if the service isn't listening then the firewall shouldn't allow it.

tasks/firewall.yml
mattwillsher: It should be the same unless there is some Ansible internal reason why not. I'd prefer to see it as a number through out, if possible.

mattwillsher: Minor style point - is [22] above, [ 22 ] here.

tasks/selinux.yml
```diff
@@ -0,0 +1,18 @@
...
+      - ports: "{{ sshd_item }}"
+        proto: tcp
+        setype: ssh_port_t
+        state: present
+        local: true
+  loop: "{{ __sshd_ports_from_config | from_json | d([]) }}"
+  loop_control:
+    loop_var: sshd_item  # avoid conflicts with the selinux loops
+  when:
+    - sshd_manage_selinux | bool
```
mattwillsher, line 16: Should this conditional be on the include_task call? Maybe the others too as it makes it quicker for some one who is browsing code to see the condition, plus there are already conditions on that call.

meta/collection-requirements.yml
mattwillsher: Manual installation isn't ideal but that's an Ansible issue we're not going to fix here.

mattwillsher:
> we want to use Ansible constructs like include_role to "call" roles only where we want, and conditionally, depending on dynamic conditions using when

I'm still wary about having cross cutting operations like this, particularly where than are OS dependent and only being implemented for one OS.  It creates difference implementation patterns for different OSs, though I understand the convenience to the role user to have it in one place. 

As there is already precedence for this in other modules, and removed a significant work burden from a the role user, I'm happy with the overall change.

tasks/selinux.yml
Jakuje: Adding them to the `install.yml` will have the advantage that the include will not be executed and the logs on unrelated platforms or when this is disabled will have one less action.
I think I kept them only in the included file, as I wanted to keep the change in the `install.yml` smaller, but I think they would be better moved to the `include_role`.

tasks/firewall.yml
Jakuje: Right. I have numbers already. For consistency, I will use the `[22]` everywhere.

README.md
Jakuje: no. This sentence is mostly about "when you change port from 22 to 222 and then from 222 to 2222 using this role, the role can not know what was the _previous_ port to remove it when adding a new one".

If somebody configures the `sshd_Port: 222` in this role, the sshd will listen only on the 222 port. What I was refering to is that if the host is managed through SSH, I do not think there is a guarantee that the following commands after the sshd restart will work (in most cases the control socket will be used, but for new reconnection, the control node will have to change the port anyway). Morover if we would get the previous port closed on the firewall, it would make the things even worse.

Jakuje:
> Right - if you use the standard documented role dependency interface - https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-dependencies - then the roles not only are downloaded, they are executed in the order specified. For all of the other system roles, _we do not want to do that_, we want to use Ansible constructs like `include_role` to "call" roles only where we want, and conditionally, depending on dynamic conditions using `when`. We cannot do any of that with the standard role dependency mechanism, which is why we invented our own that @Jakuje linked below.
> 
> > I see this information missing from https://linux-system-roles.github.io/documentation/role-requirements.html but I know it was mentioned somewhere internally.
> 
> Which information? The information about the standard Ansible role dependency mechanism?

Yeah, mostly the background why the `dependencies` in `meta/main.yml` is not used, because that would be the first question of anyone familiar with this concept

richm:
> > Right - if you use the standard documented role dependency interface - https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-dependencies - then the roles not only are downloaded, they are executed in the order specified. For all of the other system roles, _we do not want to do that_, we want to use Ansible constructs like `include_role` to "call" roles only where we want, and conditionally, depending on dynamic conditions using `when`. We cannot do any of that with the standard role dependency mechanism, which is why we invented our own that @Jakuje linked below.
> > > I see this information missing from https://linux-system-roles.github.io/documentation/role-requirements.html but I know it was mentioned somewhere internally.
> > 
> > 
> > Which information? The information about the standard Ansible role dependency mechanism?
> 
> Yeah, mostly the background why the `dependencies` in `meta/main.yml` is not used, because that would be the first question of anyone familiar with this concept

ok - if you want to submit a PR for https://github.com/linux-system-roles/linux-system-roles.github.io/blob/master/documentation/role-requirements.md I will approve it

tasks/selinux.yml
```diff
@@ -0,0 +1,16 @@
...
+- name: Ensure the custom ports are configured in selinux
+  ansible.builtin.include_role:
+    name: fedora.linux_system_roles.selinux
+  vars:
+    selinux_ports:
+      - ports: "{{ sshd_item }}"
+        proto: tcp
+        setype: ssh_port_t
+        state: present
+        local: true
```
Jakuje, line 11: reading this once again, I am not completely sure from the selinux documentation if the `local: true` should be here or not and what consequences are coming from that. @nhosoi can you have a look? In any case, the same value should be in the test when restoring the values/removing these modifications.

Jakuje: Thanks for clarification. Sounds reasonable.

Jakuje:
With the last push I fixed to test to remove the modification done by the role. Now, I think we should be good to go, unless there are some other concerns.

richm (approved):

Jakuje:
@mattwillsher do you have some more questions/comments/concerns or is this good to go?

mattwillsher:
@Jakuje I'm happy for it to be merged.

Jakuje:
Thanks! Merging.

PR #217, opened 2023-02-17
Add vars file for SUSE

nbhatia-yb:
Adding vars file for SUSE based on the contents from

https://github.com/willshersystems/ansible-sshd/blob/master/vars/Suse.yml


nbhatia-yb:
Closing

PR #222, opened 2023-04-10
ansible-lint - align with current Ansible recommendations

richm:
Use `true/false` instead of `yes/no`
Ensure use of FQCN for builtin modules
Use correct spacing in Jinja expressions
All tasks and plays must have a `name`, and the `name` string must begin with an uppercase letter
Use `ansible.posix.mount` instead of `ansible.builtin.mount`
Use `set -o pipefail` with `shell` module where supported by the platform shell


vars/AIX.yml
```diff
@@ -7,8 +7,8 @@ __sshd_sftp_server: /usr/sbin/sftp-server
...
 __sshd_defaults:
   Subsystem: "sftp {{ __sshd_sftp_server }}"
-__sshd_os_supported: yes
+__sshd_os_supported: true
 
-sshd_install_service: no
-sshd_manage_service: no
-sshd_allow_reload: yes
+sshd_install_service: false
+sshd_manage_service: falselse
```
mattwillsher, line 13:
```suggestion
sshd_manage_service: false
```

mattwillsher (changes requested):
One comment

mattwillsher (approved):
LGTM 👍

Jakuje:
Not huge fan of the changes of all "yes" and "no" in OpenSSH options, as some of them are really tri-state in sshd_config so using "true"/"false" instead of "yes"/"no" might be confusing for users. But technically should keep working.

richm:
> Not huge fan of the changes of all "yes" and "no" in OpenSSH options, as some of them are really tri-state in sshd_config so using "true"/"false" instead of "yes"/"no" might be confusing for users. But technically should keep working.

If you have something like this in YAML:
```yaml
ssh_X11Forwarding: yes
```
the YAML parser will internally turn the `yes` into a python `True` value - the original `yes` value is never used after that.

I agree that it may be confusing to users since all of the options in the ssh config files use the strings `"yes"` and `"no"`, but according to https://github.com/redhat-cop/automation-good-practices/tree/main/coding_style "Use true and false for boolean values in playbooks":
```
[YAML 1.1](https://yaml.org/type/bool.html) allows all variants whereas [YAML 1.2](https://yaml.org/spec/1.2/spec.html#id2803629) allows only true/false, and we want to be ready for when it becomes the default, and avoid a massive migration effort.
```
So one way to do it would be to require users who want to use yes/no to explicitly use the string values - so
```yaml
ssh_X11Forwarding: "yes"
```
but that also isn't ideal for users who have to remember to use quoting every time.

Perhaps there is a way to disable the yaml/ansible yes/no -> boolean True/False parsing?

Jakuje:
I was also thinking about the usage of the quotes, but I agree that it would be too cumbersome for users. And given that the ansible/yaml linting is going in this direction allowing/recommending only true/false values instead of yes/no, I think the least friction would be to follow the suit as you did here. I think the readme already mentions something about this case:
https://github.com/willshersystems/ansible-sshd/blob/master/README.md?plain=1#L152

So I think we are good.

PR #223, opened 2023-04-10
add dependabot check for github action updates

richm:


Jakuje:
Is there some more information how the dependabot works for the gh actions? Can it track also some ansible releases?

richm:
> Is there some more information how the dependabot works for the gh actions? Can it track also some ansible releases?

I'm not sure.  We are using it in linux-system-roles - for example, this is what it does - it opens PRs for component version upgrades like https://github.com/linux-system-roles/ssh/pull/73

mattwillsher (approved):

PR #233, opened 2023-04-26
test: skip selinux or firewall role test where not supported

richm:
Do not test with selinux or firewall if the platform does not support
those roles.


mattwillsher (approved):
LGTM

PR #239, opened 2023-06-13
chore: moved debian 7 (wheezy) config to explicit file

domrim:
This removes the `defaults/Debian.yml` file and moves it to the `defaults/Debian_7.yml` file. This prohibits rolling out ancient config on new Debian-Systems which aren't supported by this role.


Jakuje (approved):

Jakuje:
Thank you! This certainly makes sense!

PR #244, opened 2023-08-05
fix: Support inject_facts_as_vars = false

maage:
Enhancement:

Support `inject_facts_as_vars = false` in ansible.cfg.

The setting is considered safer because a compromised host cannot inject facts into variables.

Reason:

Minor security enhancement.

This setting is also recommended in some tuning guides like
https://docs.openstack.org/kolla-ansible/wallaby/user/ansible-tuning.html#fact-variable-injection
and issue mitigation guides:
https://docs.ansible.com/ansible/latest/reference_appendices/faq.html#when-is-it-unsafe-to-bulk-set-task-arguments-from-a-variable

`ansible_facts` are used only with one name. Previously for example `ansible_facts['os_family']` was also used as `ansible_os_family`. This helps maintainability.

Result:

Support `inject_facts_as_vars = false`. If setting is `true`, situation still works as expected.

Also drop `ansible` prefix from local variables to avoid possible conflicts in namespace and avoid possible confusion.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -6,10 +6,10 @@
     state: reloaded
   when:
     - sshd_allow_reload|bool
-    - ansible_virtualization_type|default(None) not in __sshd_skip_virt_env
+    - ansible_facts['virtualization_type']|default(None) not in __sshd_skip_virt_env
     - ansible_connection != 'chroot'
-    - ansible_os_family != 'AIX'
-    - ansible_os_family != 'OpenWrt'
+    - ansible_facts['os_family'] != 'AIX'
+    - ansible_facts['os_family'] != 'OpenWrt'
   listen: reload_sshd
 
 # sshd on AIX cannot be 'reloaded', it must be Stopped+Started.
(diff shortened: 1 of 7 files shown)


maage:
> Can `inject_facts_as_vars = false` be set during tests? It seems like it would be desirable.

It should be the value when testing, but I don't know how to do it just by configuring test env. For example: https://github.com/marketplace/actions/check-ansible-fedora-latest does not seem to have any way to modify the test env. I am probably missing something vital info.

I created PR for this, https://github.com/roles-ansible/check-ansible-fedora-latest-action/pull/4

If `inject_facts_as_vars = false` there can be this kind of errors. If `inject_facts_as_vars = true`, then either way to implement things should work.

I also noticed issue with my modifying script and fixed it. If value was `false` then the issue fixed would have been visible.

mattwillsher:
I haven't forgotten this - I'm currently on a break and will review this and the other PR week after next

Jakuje:
If there are no other concerns from @mattwillsher I can merge this. With https://github.com/roles-ansible/check-ansible-fedora-latest-action/pull/4 the CI fails for Fedora now.

mattwillsher:
I wanted to ensure `inject_facts_as_vars = false` during tests but otherwise this is good by me.

Jakuje:
> I wanted to ensure `inject_facts_as_vars = false` during tests but otherwise this is good by me.

We would have to updated all the test actions.

For now it fails the Fedora tests in more recent CI runs as the config is already in there:

https://github.com/willshersystems/ansible-sshd/actions/runs/5965231906/job/16182181136?pr=249

Jakuje:
Thank you for your contribution!

PR #245, opened 2023-08-05
security: use quote with command, shell and validate with variable

maage:
Enhancement:

Use `quote` always when using `command` plugin, `shell` plugin, validate  content as command to execute or system to configure.

Reject bad content for `systemd.unit` and `sysctl` config.

Reason:

This improves security and robustness. At least TMPDIR can be defined outside in various ways.

This follows example in `tests/tests_all_options.yml` where `pkg_mgr` is quoted.

When using `command` plugin, it could be possible to use `argv` functionality and avoid quoting.

Initial reason was single quotes in `'{{ sshd_test_hostkey.path }}/rsa_key'` in `tasks/install.yml` but not in `tasks/install_config.yml` and `tasks/install_namespace.yml`.

Result:

Included test passes. Care should be taken if test fails. It might leave `BADFLAG` user or related files behind.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -4,6 +4,23 @@
   when:
     - not __sshd_os_supported | bool
 
+- name: Ensure variables used in templates in shell or systemd unit contexts are okay
+  ansible.builtin.assert:
+    that:
+      - (sshd_sysconfig_use_strong_rng | string) == (sshd_sysconfig_use_strong_rng | quote)
+      - sshd_binary == (sshd_binary | quote)
+      - sshd_config_file == (sshd_config_file | quote)
+      - (__sshd_runtime_directory is none) or (__sshd_runtime_directory == (__sshd_runtime_directory | quote))
+      - __sshd_runtime_directory_mode == (__sshd_runtime_directory_mode | quote)
+    msg: |
+      sshd_sysconfig_use_strong_rng: {{ sshd_sysconfig_use_strong_rng }} == {{ sshd_sysconfig_use_strong_rng | quote }}
+      sshd_binary: {{ sshd_binary }} == {{ sshd_binary | quote }}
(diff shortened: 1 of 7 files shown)


maage:
Changed test to use flagfile instead of user.
Also fixed `assert` `fail_msg` -> `msg` to support older ansible versions.

maage:
Decided to split test backup / restore from main commit and some commit message fixes.

Jakuje:
The centos 8 fails reproducibly in your new test for some reason. Can you have a look for the reason? I see it fails on `SyntaxError`, but I did not dig deep enough to figure out what is the issue there (something on ansible side missing some sanitization?)

richm:
ping - can this be rebased?

richm:
> ping - can this be rebased?

ping ping

mattwillsher:
Any views on what to do with this? There is outstanding comment about the part of a path that gets quoted. If this is a significant improvement we can pick it up. In my mind, which quote will help avoid injection exploits Ansible 'should not' be getting run with unchecked, unsanitised variables, though maybe I'm missing something in my understanding.

richm:
> Any views on what to do with this? There is outstanding comment about the part of a path that gets quoted. If this is a significant improvement we can pick it up. In my mind, which quote will help avoid injection exploits Ansible 'should not' be getting run with unchecked, unsanitised variables, though maybe I'm missing something in my understanding.

Yes, this is a good security hardening patch.  Yes, users of the role should not be running Ansible with unchecked, unsanitised variables, but I suppose it could happen.

Jakuje:
Do either of you wants to take it into new PR as this one is stale, or should I do that?

mattwillsher:
If you have the time @Jakuje it would be appreciated otherwise I'll pick it up when I have a bit of time

Jakuje:
Rebased and moved to  #298 -- will address the CI issues if there will be some. Closing this one.

PR #249, opened 2023-08-24
Makes runtime dir relative

nkakouros:
Enhancement:
Makes systemd RuntimeDirectory service file directive relative (`sshd` instead of `/run/sshd`).

Reason:
The [docs](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=) say it has to be relative.

Result:
The following error is gone from the journal:

```
/etc/systemd/system/backdoor-ssh.service:14: RuntimeDirectory= path is not valid, ignoring assignment: /run/custom-ssh
```

Waiting for the tests.

Issue Tracker Tickets (Jira or BZ if any): none

### DIFF
--- a/vars/Debian_10.yml
+++ b/vars/Debian_10.yml
@@ -12,4 +12,4 @@ __sshd_defaults:
   AcceptEnv: LANG LC_*
   Subsystem: "sftp {{ __sshd_sftp_server }}"
 __sshd_os_supported: true
-__sshd_runtime_directory: /run/sshd
+__sshd_runtime_directory: sshd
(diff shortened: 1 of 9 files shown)


nkakouros:
The first failing test is a false positive (I git pushed -f a new commit message, the test uses the old one). 

The second test failing is not due to my PR.

mattwillsher:
Thanks for the PR. I'll look at the failing checks week after next, once back of holiday

Jakuje (approved):

Jakuje:
The Failed CI should be caused by roles-ansible/check-ansible-fedora-latest-action/pull/4 being merged before #244 so I believe this is also good to go here.

Thank you for your contribution!

PR #252, opened 2023-08-31
feat: manage ssh certificates

EmyLIEUTAUD:
**Enhancement:**
- Deploy User CA on the system
- Configure principals (optional)

**Reason:**
This allows you to configure and manage the SSH server to authenticate via certificates. 
Improves SSH authentication security: certificates have a validity period, unlike SSH keys.

More information on SSH certificates is available here: [Managing SSH Access at Scale with HashiCorp Vault](https://www.hashicorp.com/blog/managing-ssh-access-at-scale-with-hashicorp-vault).

**Result:**
All tests passed.
The related documentation is available and an example can be found in ```examples/example-use-certificates.yml```.

**Issue Tracker Tickets (Jira or BZ if any):** -


tasks/certificates.yml
```diff
@@ -0,0 +1,27 @@
...
+  ansible.builtin.file:
+    path: "{{ sshd['TrustedUserCAKeys'] | dirname }}"
+    state: directory
+    mode: '0755'
+
+- name: Copy Trusted user CA Keys
+  ansible.builtin.template:
+    src: "trusted-user-ca-keys.pem.j2"
+    dest: "{{ sshd['TrustedUserCAKeys'] }}"
+    mode: '0600'
```
Jakuje, line 12: I find it confusing that you are making a file with `.pem` extension which lists SSH keys in non-PEM format.

tests/tests_set_common.yml
```diff
@@ -18,6 +18,15 @@
...
           Subsystem: "sftp internal-sftp"
+          PasswordAuthentication: false
+          TrustedUserCAKeys: /etc/ssh/trusted-user-ca-keys.pem
+          AuthorizedPrincipalsFile: "/etc/ssh/auth_principals/%u"
+        sshd_trusted_user_ca_keys_list:
+          - ssh-rsa AAAAB3NzxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxwARd65YFPTgb6xxbaIB4z9Vw==
+        # Key is the user in the os, values are *Principals* defined in the certificate
+        sshd_principals:
+          user:
+            - principal
```
Jakuje, line 29: please add asserts below to make sure the options are propagated to the configuration option and make sure the needed files are created.

defaults/main.yml
```diff
@@ -44,6 +44,12 @@ sshd: {}
 # configuration file snippet or configuring second sshd service
 sshd_config_file: "{{ __sshd_config_file }}"
 
+# If not null, list of trusted CA keys
+sshd_trusted_user_ca_keys_list: []
```
Jakuje, line 48: comment does not math the default value here. Probably mention empty instead of null in the comment?

tasks/certificates.yml
EmyLIEUTAUD: I used the extension given as an example in the HashiCorp tutorial. In any case, the extension doesn't matter. Do you know a format that would be suitable without being confusing?

tests/tests_set_common.yml
Jakuje: Thanks. Given how simple the templates are, I do not think we need to check the content.

tasks/certificates.yml
Jakuje: Generally the SSH public keys have the `.pub` extension so I would probably go with that one.

templates/auth_principals.j2
```diff
@@ -0,0 +1,3 @@
+{% for principal in item.value %}
```
spetrosi, line 1:
```suggestion
{{ ansible_managed | comment }}
{% for principal in item.value %}
```
It's a good practice to add `{{ ansible_managed | comment }}` for the config file templates to notify people looking into it that this file is managed by ansible and that manual edits would be overwritten.
Is it a good idea to add it here?

templates/trusted-user-ca-keys.pub.j2
```diff
@@ -0,0 +1,3 @@
+{% for key in sshd_trusted_user_ca_keys_list %}
```
spetrosi, line 1:
```suggestion
{{ ansible_managed | comment }}
{% for key in sshd_trusted_user_ca_keys_list %}
```

tasks/certificates.yml
```diff
@@ -0,0 +1,27 @@
...
+    state: directory
+    mode: '0755'
+
+- name: Copy Trusted user CA Keys
+  ansible.builtin.template:
+    src: "trusted-user-ca-keys.pub.j2"
+    dest: "{{ sshd['TrustedUserCAKeys'] }}"
+    mode: '0600'
+
+- name: Create Pincipals directory
```
mattwillsher, line 14:
```suggestion
- name: Create Principals directory
```

mattwillsher (changes requested):
One spelling correctiong but otherwise lgtm

examples/example-use-certificates.yml
```diff
@@ -0,0 +1,23 @@
...
+        name: ansible-sshd
+      vars:
+        sshd:
+          # Disable password authentication, use SSH Certificates and configure authorized principals
+          PasswordAuthentication: false
+          TrustedUserCAKeys: /etc/ssh/trusted-user-ca-keys.pub
+          AuthorizedPrincipalsFile: "/etc/ssh/auth_principals/%u"
+        # List of trusted user CA keys
+        sshd_trusted_user_ca_keys_list:
+          - ssh-rsa AAAAB3Nzxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxd65YFPTgb6xxbaIB4z9Vw==
```
Jakuje, line 16: I would propose to put here some real key instead of the xxxs. It does not have to be your key, just generate random ssh key with ssh-keygen and put a public key here. I would suggest to use ecdsa or ed25519 as they are shorter.

tasks/certificates.yml
```diff
@@ -0,0 +1,27 @@
...
+- name: Create Trusted user CA Keys directory
+  ansible.builtin.file:
+    path: "{{ sshd['TrustedUserCAKeys'] | dirname }}"
+    state: directory
+    mode: '0755'
+
+- name: Copy Trusted user CA Keys
+  ansible.builtin.template:
+    src: "trusted-user-ca-keys.pub.j2"
+    dest: "{{ sshd['TrustedUserCAKeys'] }}"
```
Jakuje, line 11: All of the configuration options we have in this role work in both writing as `sshd['TrustedUserCAKeys']` as well as `sshd_TrustedUserCAKeys` so I think this part of the role should reflect this too. I was solving similar thing with hostkeys here:

https://github.com/willshersystems/ansible-sshd/blob/d54f51f32a34cb0c08c722c944dc616a3b1c1bcd/tasks/install.yml#L37-L52

The same applies for `AuthorizedPrincipalsFile` below.

Jakuje (changes requested):

tasks/certificates.yml
EmyLIEUTAUD: Should I add the to_json filter to maintain Python 2 compatibility?

Jakuje: Yes please. I hope we will not need it but better be safe than sorry.

tasks/certificates.yml
```diff
@@ -0,0 +1,27 @@
...
+    state: directory
+    mode: '0755'
+  when: sshd_principals != {}
+
+- name: Copy Principals files
+  ansible.builtin.template:
+    src: "auth_principals.j2"
+    dest: "{{ sshd['AuthorizedPrincipalsFile'] | dirname }}/{{ item.key }}"
+    mode: '0644'
+  loop: "{{ q('dict', sshd_principals) }}"
```
spetrosi, line 26:
```suggestion
  loop: "{{ ('dict', sshd_principals) }}"
```
Is it a typo?

tests/tests_set_common.yml
```diff
@@ -52,6 +61,29 @@
...
+              - trustedusercakeys_file_stat.stat.exists
+
+        - name: Get AuthorizedPrincipalsFile stat
+          ansible.builtin.stat:
+            path: /etc/ssh/auth_principals/user
+          register: authorizedprincipalsfile_stat
+
+        - name: Check AuthorizedPrincipalsFile exists
+          ansible.builtin.assert:
+            that:
```
spetrosi, line 85:
```suggestion
            that: authorizedprincipalsfile_stat.stat.exists
```

tests/tests_set_common.yml
```diff
@@ -52,6 +61,29 @@
...
+              - authorizedprincipalsfile_stat.stat.exists
```
spetrosi, line 86:
```suggestion
```

tests/tests_set_common.yml
```diff
@@ -52,6 +61,29 @@
...
+              - "'AuthorizedPrincipalsFile /etc/ssh/auth_principals/%u' in config.content | b64decode"
+
+        - name: Get TrustedUserCAKeys file stat
+          ansible.builtin.stat:
+            path: /etc/ssh/trusted-user-ca-keys.pub
+          register: trustedusercakeys_file_stat
+
+        - name: Check TrustedUserCAKeys file exists
+          ansible.builtin.assert:
+            that:
```
spetrosi, line 75:
```suggestion
            that: trustedusercakeys_file_stat.stat.exists
```

tests/tests_set_common.yml
```diff
@@ -52,6 +61,29 @@
...
+              - trustedusercakeys_file_stat.stat.exists
```
spetrosi, line 76:
```suggestion
```

tasks/certificates.yml
EmyLIEUTAUD: Changes added: [d301935](https://github.com/willshersystems/ansible-sshd/pull/252/commits/d3019356d3d703906dbad8700319580d7c3ebdf9)

tasks/certificates.yml
EmyLIEUTAUD: This was an old syntax we used, but it's no longer documented. I've modified it to make the code easier to understand.

tasks/certificates.yml
```diff
@@ -0,0 +1,46 @@
...
+        {{ sshd_TrustedUserCAKeys | to_json }}
+      {% else %}
+        {{ sshd['TrustedUserCAKeys'] | to_json }}
+      {% endif %}
+  block:
+    - name: Create Trusted user CA Keys directory
+      ansible.builtin.file:
+        path: "{{ (__sshd_trustedusercakeys_from_config | from_json) | dirname }}"
+        state: directory
+        mode: '0755'
```
richm, line 16: If the directory already exists and has a more secure mode e.g. `'0700'` do you want this to change it?

README.md
```diff
@@ -282,6 +282,44 @@ Default path to the sftp server binary.
...
+    - some-principal-defined-in-certificate
+```
+
+### Configuration needed
+
+We need to configure the paths to principals (optional) and the trusted CA keys file:
+
+```yaml
+sshd:
+  PasswordAuthentication: false
```
Jakuje, line 316: I would leave out this line as it is not required for the certificate authentication (even though it might be a common to disable the passwords when deploying certificates).

Jakuje (commented):
hope last nits to the wording of README. Then I would love to have the fixup commits slightly squashed as the commit series is getting long ...

tasks/certificates.yml
richm: > Would you prefer to not specify the mode and use the existing mode if the file exists, and otherwise use the default mode? Or should I add a task to check if the directory already exists, and change mode only if it's less secure?

We typically do the latter in other system roles (do a `stat`, then use `stat.mode` if `stat.exists`, else `0755`), but if `0755` is the default for sshd, and recommended, then I guess it is ok

mattwillsher: Fwiw I think the role should declare it in line with good practice. 

Using what exists in my view can lead to spaghetti dependencies and unexpected outcomes. Let this role own the directory or else let something else take responsibility.

If there is variation across releases a distribution specific variable can be used with an override, common to how this is handled for other variant cases. A variable seems sensible here as it could be considered security sensitive to some.

README.md
```diff
@@ -282,6 +282,44 @@ Default path to the sftp server binary.
...
+  TrustedUserCAKeys: /etc/ssh/path-to-trusted-user-ca-keys/trusted-user-ca-keys.pub
+  AuthorizedPrincipalsFile: "/etc/ssh/path-to-auth-principals/auth_principals/%u"
+```
+
+To learn more about SSH Certificate, here is a [nice tutorial from Hashicorp](https://www.hashicorp.com/blog/managing-ssh-access-at-scale-with-hashicorp-vault)
```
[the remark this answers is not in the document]
EmyLIEUTAUD: I think the first tutorial is good because we can see how to set up SSH user certificates. 

What do you think about also leaving the HashiCorp tutorial to explain the principles part? I think it's useful for understanding what they're for and how they can be set up. They are available in the open-source version of Vault.

tasks/certificates.yml
EmyLIEUTAUD: If I've understood correctly, I'll need to create a dedicated role for certificates so that the directories in question are no longer owned by root:root?

.pre-commit-config.yaml
```diff
@@ -8,7 +8,7 @@ repos:
         types: [file, yaml]
         entry: yamllint --strict
   - repo: https://github.com/ansible/ansible-lint.git
-    rev: v6.5.2
+    rev: v6.17.2
```
Jakuje, line 11: please, keep the dependencies changes in at least separate commit (ideally separate PR).

README.md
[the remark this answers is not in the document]
Jakuje: I am ok with both of the links. Thanks for the patience!

tests/tests_set_common.yml
```diff
@@ -52,6 +61,27 @@
...
+            that: trustedusercakeys_file_stat.stat.exists
+
+        - name: Get AuthorizedPrincipalsFile stat
+          ansible.builtin.stat:
+            path: /etc/ssh/auth_principals/user
+          register: authorizedprincipalsfile_stat
+
+        - name: Check AuthorizedPrincipalsFile exists
+          ansible.builtin.assert:
+            that: authorizedprincipalsfile_stat.stat.exists
```
Jakuje, line 84: Given that we have now option to set also the modes/owners of the files, I would check it also here in the test.

But given that this test is growing out of the scope a bit, I would probably propose to split out the certificate to separate test (and thoughtfully test also the new owner/group/mode options that were recently introduced  -- it should not be that much code given that you already do the stat here.

.pre-commit-config.yaml
EmyLIEUTAUD: Version restored here: [610a356](https://github.com/willshersystems/ansible-sshd/pull/252/commits/610a356d48ac8089ed374937350b414dbe68461c]).

README.md
```diff
@@ -282,6 +282,62 @@ Default path to the sftp server binary.
...
+
+### Additional variables
+
+#### sshd_trusted_user_ca_keys_list
+
+List of the trusted user CA public keys in OpenSSH (one-line) format (mandatory).
+
+#### sshd_trustedusercakeys_directory_owner, shsd_trustedusercakeys_directory_group, sshd_trustedusercakeys_directory_mode
+
+Use these variables to set the ownership and permissions for the Trusted User CA Keys directory. Defaults are respectively *0750*, *root* and *root*.
```
mattwillsher, line 300: Default example order should have mode last, as with the other 3 cases

mattwillsher (changes requested):
Minor ordering in docs but looks good to be otherwise.

I don't understand why the checkout action is showing the version change - have you rebased from main recently? It might resolve that

mattwillsher (approved):

richm (approved):

tasks/certificates.yml
EmyLIEUTAUD: By default, I've set the trusted user ca keys directory to 0755, since in tutorials this is often the ssh directory and if it's set to 0750 we can't connect to the machine.

Jakuje (approved):
code wise it looks good. I would prefer the fixups commits squashed, but if others are ok with the commit series. I do not want to block the changes

mattwillsher:
Thanks for all the work on this @EmyLIEUTAUD 

I'll get a new release out that includes this code in the few days.

PR #251, opened 2023-08-31
docs(changelog): version v0.21.0

mattwillsher:
Change log for 0.21.0

PR #255, opened 2023-09-12
docs(changelog): version v0.21.0 [citest skip]

spetrosi:
Update changelog and .README.html for version v0.21.0

### DIFF
--- a/.README.html
+++ b/.README.html
@@ -1,0 +1,678 @@
+<!DOCTYPE html>
+<!--
+==============================================================================
+           "GitHub HTML5 Pandoc Template" v2.2 — by Tristano Ajmone
+==============================================================================
+Copyright © Tristano Ajmone, 2017-2020, MIT License (MIT). Project's home:
+
+- https://github.com/tajmone/pandoc-goodies
+
+The CSS in this template reuses source code taken from the following projects:
+
+- GitHub Markdown CSS: Copyright © Sindre Sorhus, MIT License (MIT):
+  https://github.com/sindresorhus/github-markdown-css
+
+- Primer CSS: Copyright © 2016-2017 GitHub Inc., MIT License (MIT):
+  http://primercss.io/
+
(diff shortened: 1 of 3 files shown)


spetrosi:
@mattwillsher opening this instead of  #251 to add recent commits and .README.html. We use .README.html when building RPM. We used to convert md>html during the build, but the build environment has limited packages, it became hard to properly convert into a pretty HTML there.

mattwillsher:
How was the .README.html generated @spetrosi ?

spetrosi:
> How was the .README.html generated @spetrosi ?

The release script gets README.html from origin/docs:latest/README.html, see https://github.com/linux-system-roles/auto-maintenance/blob/main/role-make-version-changelog.sh#L309C18-L309C18
And README in the docs branch is generated by https://github.com/willshersystems/ansible-sshd/blob/main/.github/workflows/build_docs.yml

mattwillsher:
We need a doc detailing the process here if it's more that just updating the CHANGELOG.md. 

I'm unclear why .README.html shows in this single commit if it was generated by some other script. 

The process needs to be straight forward for any one who is doing a release.

spetrosi:
I agree, it's complicated now, I'll add documentation to auto-maintenance repo. I had an idea to create a github action that anyone can run ad hoc to create a new release but didn't get to it yet.
To your question, the latest/README.html from docs branch is copied to the main branch `.README.html` so that we can get it in RPM spec during build without the need to pull one more source, which would need to be the docs branch in addition to the main branch for each repo.
And the script copies to a  hidden .README.html because it would get outdated between the releases when new changes to README.md are done. So this hidden .README.hml should not be used actually.

mattwillsher:
@spetrosi - please add the steps as a list of commands and manual steps, runbook style, to the README.md in a new section near the end of the document. I would like to ensure releases are consistent regardless of who is performing the release.

spetrosi:
I've added a contributing.md file to keep README.md targeted to end-users only. Does this work?

mattwillsher:
Brilliant, thats even better, thank you @spetrosi

mattwillsher (approved):

mattwillsher:
This is now released - the old role in galaxy needed to be deleted due to the change in name of the default branch.

PR #257, opened 2023-09-22
Support socket activation with non-default ports

eliasp:
Enhancement:
Control via the var `sshd_socket_activation` whether SSH should run as a single permanent service or whether on each incoming connection request, a new instance of `sshd@.service` should be spawned.

Reason:
This improves security, allows for easier per-connection troubleshooting and eliminates the need to restart the service after config changes.

Result:
- `sshd.socket` is running, `sshd.service` is not
- on each connection, a service instance like `sshd@1-10.13.13.5:22-192.168.178.53:44876.service` is spawned

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,12 @@ Using these variables, you can use your own custom templates. With the above
 default templates, the name of the installed ssh service will be provided by
 the `sshd_service` variable.
 
+#### sshd_socket_activation
+
+If set to *true*, systemd's socket activation will be used to start service
+instances (`sshd.socket` activates `sshd@.service`) on each incoming connection
+instead running a single permanent service for all connections.
+
 #### sshd_manage_firewall
 
 If set to *true*, the the SSH port(s) will be opened in firewall. Note, this
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -2,7 +2,7 @@
 
 - name: Reload the SSH service
   ansible.builtin.service:
-    name: "{{ sshd_service }}"
+    name: "{{ sshd_service }}{{ '.socket' if sshd_socket_activation | bool else '' }}"
     state: reloaded
   when:
     - sshd_allow_reload|bool
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,4 +1,12 @@
 ---
+- name: Reload systemd units
+  ansible.builtin.systemd:
+    daemon_reload: true
+  when:
+    - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
+    - ansible_connection != 'chroot'
+    - ansible_facts['os_family'] not in ['AIX', 'OpenWrt']
+  listen: reload_units
 
 - name: Reload the SSH service
   ansible.builtin.service:
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -13,6 +13,10 @@ sshd_manage_service: true
 # If the below is false, don't reload the ssh daemon on change
 sshd_allow_reload: true
 
+# If set to true, use systemd's socket activation and start a socket unit,
+# instead of the service unit
+sshd_socket_activation: "{{ __sshd_socket_activation | bool }}"
+
 # If the below is true, also install service files from the templates pointed
 # to by the `sshd_service_template_*` variables
 sshd_install_service: false
--- a/tasks/install_service.yml
+++ b/tasks/install_service.yml
@@ -29,7 +29,7 @@
 
 - name: Service enabled and running
   ansible.builtin.service:
-    name: "{{ sshd_service }}"
+    name: "{{ sshd_service }}{{ '.socket' if sshd_socket_activation | bool else '' }}"
     enabled: true
     state: started
   when:
--- a/tasks/install_service.yml
+++ b/tasks/install_service.yml
@@ -1,31 +1,34 @@
 ---
-- name: Install systemd service files
+- name: Install service unit file
+  ansible.builtin.template:
+    src: "{{ sshd_service_template_service }}"
+    dest: "/etc/systemd/system/{{ sshd_service }}.service"
+    owner: root
+    group: root
+    mode: "0644"
+  notify: reload_sshd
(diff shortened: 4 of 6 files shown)


tasks/install_config.yml
```diff
@@ -24,6 +24,7 @@
         {{ sshd_binary }} -t -f %s
       {% endif %}
     backup: "{{ sshd_backup }}"
+  changed_when: not sshd_socket_activation|bool
```
mattwillsher, line 27: Why is this set?
I understand that this will cause this block to appear changed every run when `sshd_socket_activation` is set `false`

README.md
```diff
@@ -109,6 +109,12 @@ Using these variables, you can use your own custom templates. With the above
 default templates, the name of the installed ssh service will be provided by
 the `sshd_service` variable.
 
+#### sshd_socket_activation
+
+If set to *true*, systemd's socket activation will be used to start service
+instances (`sshd.socket` activates `sshd@.service`) on each incoming connection
```
mattwillsher, line 115: I don't see a reference to `ssh@.service` - this already exists? Is implicite by the name of the service?

mattwillsher (commented):
Thank you for the PR. I've added a couple of comments.

Is there anything else that would need to be added to ensure complete managements of socket base instantiation? Ubuntu 22.10 appears to use sockets by default. Do you know if this change will fully support that implementation?

tasks/install_config.yml
eliasp: With a conventional service, a reload is needed on config changes.
With a socket activated service, no reload is needed, since the config will be read as soon as a new connection spawns a new service instance.

So don't emit any changes here to prevent a service reload.

README.md
eliasp: No reference to `ssh@.service` is needed, that's automatically derived (unless specified differently in the socket unit) by systemd from the socket unit's name. E.g.
- `sshd.socket` > `sshd@.service`
- `foo-bar.socket` > `foo-bar@.service`
- ...

eliasp:
> Thank you for the PR. I've added a couple of comments.
> 
> Is there anything else that would need to be added to ensure complete managements of socket base instantiation? Ubuntu 22.10 appears to use sockets by default. Do you know if this change will fully support that implementation?

I can give Ubuntu 22.10 a try, but in general, I don't see why this shouldn't work. What should probably be done is to set the OS specific defaults for `sshd_socket_activation` accordingly.

I have pushed a few more commits to handle non-default SSH ports properly and converted the PR to a Draft for now, since I think a few more changes might be needed:
- handle a reset of `ansible_port` after the service moves to a different port while the play is running (I already have a nice idea how to handle this quite smoothly, but I need to give it a try)
- I need to ensure non-standard ports also work as expected with non-socket-activated setups
- I discovered some oddities regarding the config path in the service units with `sshd_skip_defaults: true` on RHEL-based systems.

handlers/main.yml
```diff
@@ -1,8 +1,17 @@
 ---
+- name: Reload systemd units
+  ansible.builtin.systemd:
+    daemon_reload: true
+  when:
+    - ansible_facts['virtualization_type']|default(None) not in __sshd_skip_virt_env
```
richm, line 6:
```suggestion
    - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
```

handlers/main.yml
```diff
@@ -1,8 +1,17 @@
...
+    - ansible_connection != 'chroot'
+    - ansible_facts['os_family'] != 'AIX'
```
richm, line 8:
```suggestion
    - ansible_facts['os_family'] not in ['AIX', 'OpenWrt']
```

tasks/install_config.yml
richm, line 27:
```suggestion
  changed_when: not sshd_socket_activation | bool
```

tasks/install_config.yml
```diff
@@ -43,6 +44,7 @@
...
```
richm, line 47:
```suggestion
  changed_when: not sshd_socket_activation | bool
```

tasks/install_namespace.yml
```diff
@@ -21,4 +21,5 @@
         {{ sshd_binary }} -t -f %s
       {% endif %}
     backup: "{{ sshd_backup }}"
+  changed_when: not sshd_socket_activation|bool
```
richm, line 24:
```suggestion
  changed_when: not sshd_socket_activation | bool
```

tasks/install_config.yml
eliasp: I dropped the `changed_when` params, since I misunderstood them.
I understood them as a filter, so that a change from this task is only reported, when the defined condition is met.
Instead, it acts as a general condition to always report a change when this condition is met.

handlers/main.yml
```diff
@@ -1,8 +1,16 @@
...
+  when:
+    - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
+    - ansible_connection != 'chroot'
+    - ansible_facts['os_family'] not in ['AIX', 'OpenWrt']
+  listen: reload_units
 
 - name: Reload the SSH service
   ansible.builtin.service:
-    name: "{{ sshd_service }}"
+    name: "{{ sshd_service }}{{ '.socket' if sshd_socket_activation | bool else '' }}"
```
Jakuje, line 13: nit: I would propose to use temporary variable for this as it is not visible on the first sight what is being done. Something like
```
__sshd_systemd_unit: "{{ sshd_service }}{{ '.socket' if sshd_socket_activation | bool else '' }}"
```
and then 
```
  ansible.builtin.service:
    name: "{{ __sshd_systemd_unit }}"
```
The task name no longer matches description either. It should be `Reload the SSH unit` or something like that.

richm:
> ping - any update?

ping ping

mattwillsher:
Give the outstanding comments I'll close this off in a week or two if there are no further updates.

mattwillsher:
Closed due to inactivity

Issue #264, opened 2023-10-18
Role creates /root/sshd directory on debian


Yannik:
On debian (10/11/12), the `vars` file contains this line:

```
__sshd_runtime_directory: sshd
```

In `install.yml`, the following task is executed:

```
    - name: Make sure sshd runtime directory is present
      ansible.builtin.file:
        path: "{{ __sshd_runtime_directory }}"
        state: directory
        owner: root
        group: root
        mode: "{{ __sshd_runtime_directory_mode }}"
      when:
        - __sshd_runtime_directory is not none
```

This results in an directory `/root/sshd` being created. (Assuming ansible_user is root)

Obviously, this serves no purpose and should not be done.

Jakuje:
Thanks for the report!

This was caused by some refactoring. When this task was introduced in 6b36488299e742138022c745a948d7c29e88ae1e the `__sshd_runtime_directory` was pointing to `/run/sshd`, which is normally created by systemd when starting the service and which is needed for the `validate` check to pass. But d6abe9c0325183180e6f83ba14c28781b9e6c6b9 changed the paths to relative (dropping the `/run/` prefix), which caused this bug. This should be fixed with #265

Yannik:
@Jakuje Great, thanks. Looking forwarding to seeing #265 merged.

Jakuje:
Merged as #265. We should get a next release soon. Closing.

PR #265, opened 2023-10-23
fix: Avoid creation of runtime directories in home

Jakuje:
Enhancement: Avoid creation of runtime directory in home

Reason: The d6abe9c0325183180e6f83ba14c28781b9e6c6b9 changed the directories to be relative, but did not update tasks

Result: The playbook should create the runtime directory in the right place.

Issue Tracker Tickets (Jira or BZ if any): #264

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -113,7 +113,7 @@
 
     - name: Make sure sshd runtime directory is present
       ansible.builtin.file:
-        path: "{{ __sshd_runtime_directory }}"
+        path: "/run/{{ __sshd_runtime_directory }}"
         state: directory
         owner: root
         group: root
--- a/tests/tests_runtime_directory.yml
+++ b/tests/tests_runtime_directory.yml
@@ -1,0 +1,52 @@
+---
+- name: Test runtime directory
+  hosts: all
+  vars:
+    __sshd_test_backup_files:
+      - /etc/ssh/sshd_config
+      - /etc/ssh/sshd_config.d/00-ansible_system_role.conf
+  tasks:
+    - name: Backup configuration files
+      ansible.builtin.include_tasks: tasks/backup.yml
+
+    - name: Configure sshd with default options
+      ansible.builtin.include_role:
+        name: ansible-sshd
+
+    - name: Verify the runtime diretory is created on Debian
+      tags: tests::verify
+      when:
+        - ansible_facts['os_family'] == "Debian"
+      block:
+        - name: Stat the default runtime directory
+          ansible.builtin.stat:
+            path: /run/sshd
+          register: run_stat
+
+        - name: Check the runtime directory is present
+          ansible.builtin.assert:
+            that:
+              - run_stat.stat.isdir
+              - run_stat.stat.mode == "0755"
+
+    - name: Verify the runtime diretory is not created in wrong places
+      tags: tests::verify
+      block:
+        - name: Stat the home directory for the runtime directory
+          ansible.builtin.stat:
+            path: ~/sshd
+          register: home_stat
+
+        - name: Stat the /run/~ for the runtime directory
+          ansible.builtin.stat:
+            path: /run/~
+          register: run_user_stat
+
+        - name: Check the wrong runtime directory is not present
+          ansible.builtin.assert:
+            that:
+              - not home_stat.stat.exists
+              - not run_user_stat.stat.exists
+
+    - name: "Restore configuration files"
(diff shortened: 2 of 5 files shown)


tasks/install.yml
```diff
@@ -113,7 +113,7 @@
 
     - name: Make sure sshd runtime directory is present
       ansible.builtin.file:
-        path: "{{ __sshd_runtime_directory }}"
+        path: "/run/{{ __sshd_runtime_directory }}"
```
richm, line 116: Is it ok to do this for all platforms, not just Ubuntu/Debian?

Jakuje: The Fedora/RHEL do not have RuntimeDirectory in the service file and therefore they have the `__sshd_runtime_directory` empty, which should skip this task.

richm (approved):

tasks/install.yml
mattwillsher: According to 'Table 2. Automatic directory creation and environment variables' at https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html the value is always relative. 
This value is set in the unit file. 
The template for this always set the value. The default for the runtime directory is ~, which YAML should interpret as null. I don't know what Ansible does with it.

I suggest the unit file makes the Runtime values conditional on the runtime directory being not null, so excluding the value entirely if not set. 

If ~ isn't being interpreted as null - the test added implies that it isn't - maybe it should be set to something that will be or just an empty string and the template set to add the RunTime... value iff the corresponding var is non-empty string.

mattwillsher (changes requested):
The systemd unit file shouldn't include Runtime values at all if they are not set/not present in the default unit files

Jakuje:
> The systemd unit file shouldn't include Runtime values at all if they are not set/not present in the default unit files

Good point. I will update the template and test to verify this.

In separate PR I will probably do some more cleanup of the service/socket files as they are quite outdated anyway and they need some improvements to match differences between distributions.

tasks/install.yml
Jakuje: My understanding and observation from the test is that the `~` is interpreted as null/none. 
I updated the test and the templates not to print these empty options.

tests/tests_runtime_directory.yml
```diff
@@ -0,0 +1,111 @@
...
+              - '"RuntimeDirectoryMode=" not in service.content | b64decode'
+              - '"RuntimeDirectory=" not in service_inst.content | b64decode'
+              - '"RuntimeDirectoryMode=" not in service_inst.content | b64decode'
+
+    - name: Verify the runtime diretory is not created in wrong places
+      tags: tests::verify
+      block:
+        - name: Stat the home directory for the runtime directory
+          ansible.builtin.stat:
+            path: ~/sshd
```
mattwillsher, line 96: This can't happen with the new code. Test is unnecessary

tests/tests_runtime_directory.yml
```diff
@@ -0,0 +1,111 @@
...
+          register: home_stat
+
+        - name: Stat the /run/~ for the runtime directory
+          ansible.builtin.stat:
+            path: /run/~
```
mattwillsher, line 101: I don't understand the need for this - if ~ is really null in Ansible this shouldn't occur. If ~ is not null in Ansible the default value should be changed

mattwillsher (approved):
See comments on tests but the code changes LGTM

tests/tests_runtime_directory.yml
Jakuje: This is the test that I introduced to really verify that the `~` is being treated specially and not as a `~` string as I could not find this anywhere explicitly mentioned. I do not think the meaning would change in the future ansible versions, but I think this wont hurt sanity-checking that unexpected directories are not created.

One though is that I should probably make sure it does not exist before invoking the role to avoid false positives if something in the system would start creating these.

tests/tests_runtime_directory.yml
Jakuje: Right. Unless somebody will change the code again and we won't notice as in the past. This is a point of regression testing, making sure we do not re-introduce the old bugs we had in the past. Indeed, new ones can always lurk into the code, but it give at least some sanity. As in the other comment, I should probably make sure the path does not exist before running the role though.

I still find them useful, but if you insist, I can remove these.

mattwillsher:
No need to change just wondering about their value. More checks is better :)

PR #267, opened 2023-10-30
fix: Review and update service units and socket unit to include distribution defaults

Jakuje:
Enhancement: Update systemd units to reflect OS defaults

Reason: The systemd units were based on some old ubuntu/debian and never updated to match current versions.

Result: The generated units include all important system defaults

### DIFF
--- a/templates/sshd.service.j2
+++ b/templates/sshd.service.j2
@@ -1,15 +1,37 @@
 [Unit]
 Description=OpenBSD Secure Shell server
+After=network.target{{ (__sshd_service_after is none) | ternary('', ' ' ~ __sshd_service_after) }}
+{% if __sshd_service_wants is string %}
+Wants={{ __sshd_service_wants }}
+{% elif __sshd_service_wants is iterable %}
+{%   for file in __sshd_service_wants %}
+Wants={{ file }}
+{%   endfor %}
+{% endif %}
 Documentation=man:sshd(8) man:sshd_config(5)
 
 [Service]
 Type=notify
+{% if __sshd_environment_file is string %}
+EnvironmentFile=-{{ __sshd_environment_file }}
+{% elif __sshd_environment_file is iterable %}
+{%   for file in __sshd_environment_file %}
+EnvironmentFile=-{{ file }}
+{%   endfor %}
+{% endif %}
 ExecStartPre={{ sshd_binary }} -t
-ExecStart={{ sshd_binary }} -D -f {{ sshd_config_file }}
+ExecStart={{ sshd_binary }} -D {{ __sshd_environment_variable }} -f {% if sshd_main_config_file is not none %}
+{{- sshd_main_config_file }}
+{% else %}
+{{- sshd_config_file }}
+{% endif %}
 ExecReload={{ sshd_binary }} -t
 ExecReload=/bin/kill -HUP $MAINPID
 KillMode=process
 Restart=on-failure
+{% if __sshd_service_restart_timeout is not none %}
+RestartSec={{ __sshd_service_restart_timeout }}
+{% endif %}
 RestartPreventExitStatus=255
 {% if __sshd_runtime_directory is not none %}
 RuntimeDirectory={{ __sshd_runtime_directory }}
@@ -18,3 +40,6 @@ RuntimeDirectoryMode={{ __sshd_runtime_directory_mode }}
 
 [Install]
 WantedBy=multi-user.target
+{% if __sshd_service_alias is not none %}
+Alias={{ __sshd_service_alias }}.service
+{% endif %}
--- a/templates/sshd.socket.j2
+++ b/templates/sshd.socket.j2
@@ -1,12 +1,19 @@
 [Unit]
 Description=OpenBSD Secure Shell server socket
 Documentation=man:sshd(8) man:sshd_config(5)
-Before={{ sshd_service }}.service
+{% if __sshd_socket_accept %}
 Conflicts={{ sshd_service }}.service
+{% else %}
+Before=sockets.target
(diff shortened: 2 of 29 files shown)


templates/sshd.service.j2
```diff
@@ -1,15 +1,37 @@
...
+{% if __sshd_environment_file is string %}
+EnvironmentFile=-{{ __sshd_environment_file }}
+{% elif __sshd_environment_file is iterable %}
+{%   for file in __sshd_environment_file %}
+EnvironmentFile=-{{ file }}
+{%   endfor %}
+{% endif %}
 ExecStartPre={{ sshd_binary }} -t
-ExecStart={{ sshd_binary }} -D -f {{ sshd_config_file }}
+ExecStart={{ sshd_binary }} -D {{ __sshd_environment_variable }} -f {% if sshd_main_config_file is not none %}
```
richm, line 23: the `-D` argument works with and without an env. var.?

vars/main.yml
```diff
@@ -63,3 +63,25 @@ __sshd_verify_hostkeys_default: []
...
+# The systemd service name alias
+__sshd_service_alias: ~
+
+# The systemd service wants directive
+__sshd_service_wants: ~
+
+# The systemd service RestartSec directive
+__sshd_service_restart_timeout: ~
+
+# The systemd socket file does not accept the conne
```
richm, line 86:
```suggestion
# The systemd socket file does not accept the connection
```

templates/sshd.socket.j2
```diff
@@ -1,12 +1,19 @@
 [Unit]
 Description=OpenBSD Secure Shell server socket
 Documentation=man:sshd(8) man:sshd_config(5)
-Before={{ sshd_service }}.service
+{% if __sshd_socket_accept is sameas true %}
```
richm, line 4:
```suggestion
{% if __sshd_socket_accept %}
```
?

tests/tests_alternative_file.yml
```diff
@@ -135,12 +135,19 @@
...
-        - name: Check Ubuntu 22 defaults are present in the first configuration file
+        - name: Check Ubuntu 20+ defaults are present in the first configuration file
           ansible.builtin.assert:
             that:
               - "'Include /etc/ssh/sshd_config.d/*.conf' in config3.content | b64decode"
-              - "'KbdInteractiveAuthentication no' in config.content | b64decode"
               - "'UsePAM yes' in config.content | b64decode"
+          when:
+            - ansible_facts['distribution'] == 'Ubuntu'
+            - ansible_facts['distribution_major_version']|int >= 20
```
richm, line 145:
```suggestion
            - ansible_facts['distribution_major_version'] | int >= 20
```

templates/sshd.service.j2
Jakuje: Yes. The `-D` is just saying the `sshd` to daemonize. The environment variable contains then any other options provided by the environment file.

tests/tests_alternative_file.yml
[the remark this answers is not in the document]
richm: Better to be consistent with the surrounding file context - not sure why ansible-lint doesn't complain about this

Jakuje:
@richm @mattwillsher is this ready to merge or is there some more work needed and I should pull the changes fixing CI into separate PR?

mattwillsher (approved):

mattwillsher:
Looks ok to me from a cursory review

templates/sshd.service.j2
```diff
@@ -1,15 +1,37 @@
 [Unit]
 Description=OpenBSD Secure Shell server
+After=network.target {{ __sshd_service_after }}
```
richm, line 3: Will `__sshd_service_after` always be set for all valid systemd platforms?

richm: That is - if it is possible it will be `None` then you will need to do something like
```
After=network.target{{ (__sshd_service_after is None) | ternary('', ' ' ~ __sshd_service_after) }}
```

tests/tests_alternative_file.yml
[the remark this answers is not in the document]
Jakuje: I changed the filter to be more systematic in this PR adding spaces where needed.

templates/sshd.service.j2
Jakuje: Right now, it is. But its better to be safe. Fixed and rebased.

Jakuje: ok, ansible does not this:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: . Could not load "None": 'None'
```
did I made anything wrong or what does ansible does not like on this expression?

richm (approved):

richm:
I guess ansible-lint still has the bug where it will not check multiline `when` for Jinja spacing . . .

templates/sshd.service.j2
Jakuje: lowercase `none` in jinja templates?

richm: https://jinja.palletsprojects.com/en/2.9.x/templates/#literals
"

Note

The special constants true, false, and none are indeed lowercase. Because that caused confusion in the past, (True used to expand to an undefined variable that was considered false), all three can now also be written in title case (True, False, and None). However, for consistency, (all Jinja identifiers are lowercase) you should use the lowercase versions.
"

Jakuje:
So it looks like its final, green and ready to merge, unless there will be some more comments.

I did similar change regarding the After keyword also in the instantiated service file.

richm:
ready to merge?

Jakuje:
Fine with me.

PR #268, opened 2023-11-17
feat: support for ostree systems

richm:
Feature: Allow running and testing the role with ostree managed nodes.

Reason: We have users who want to use the role to manage ostree
systems.

Result: Users can use the role to manage ostree managed nodes.


richm:
I know it's a big change - many, many small changes.  One thing I could do is eliminate some of the ostree packages files for platforms that aren't really supported - I don't see many people building ostree systems for Debian, Ubuntu, etc. - then add them if there is demand.

hmm - not sure how to make the ansible checks pass - looks like just having the `ansible.utils.update_fact` module in the file causes parsing of the files to fail if the collection isn't available :-(

richm:
looking for an alternate solution :-(  I don't think this will fly with legacy role users . . .

PR #270, opened 2023-11-17
feat: support for ostree systems

richm:
Feature: Allow running and testing the role with ostree managed nodes.

Reason: We have users who want to use the role to manage ostree
systems.

Result: Users can use the role to manage ostree managed nodes.


.ostree/packages-runtime-Debian-9.txt
```diff
@@ -0,0 +1,2 @@
+openssh-server
+openssh-sftp-server
```
Jakuje, line 2: One more question, the Debian/Arch/Alpine works with ostree? I thought its fedora specific ... wiki says there are other systems using, but I did not find any working Debian version (except for the existence of the debian package).

richm: > One more question, the Debian/Arch/Alpine works with ostree? I thought its fedora specific ... wiki says there are other systems using, but I did not find any working Debian version (except for the existence of the debian package).

You are correct - afaik, only `os_family == RedHat` is supported for `rpm-ostree`.  If you like, I can change the script to ignore non-RedHat family for the purpose of generating the .ostree files.

Jakuje: Yeah. I think it would be good to avoid too many new files for none-to-very-few users, who can add it later if they would need.

richm: Done in latest commit

Jakuje:
From the failing CI, the Fedora is fixed with top commit in #267, debian buster is probably just broken (can be ignored), but centos6 is something that should be fixed as part of this PR as the `man-db` requirement was introduced/changed here.

richm:
> From the failing CI, the Fedora is fixed with top commit in #267, debian buster is probably just broken (can be ignored), but centos6 is something that should be fixed as part of this PR as the `man-db` requirement was introduced/changed here.

ok - do you want to merge #267 then I can rebase?

Jakuje:
> > From the failing CI, the Fedora is fixed with top commit in #267, debian buster is probably just broken (can be ignored), but centos6 is something that should be fixed as part of this PR as the `man-db` requirement was introduced/changed here.
> 
> ok - do you want to merge #267 then I can rebase?

Whatever works for you. The #267 changes a lot of things, but they are not used by default so they should not be breaking (but some careful review would be needed). It does not have any deadlines as far as I know.

If the ostree support needs to land at some reasonable time, feel free to pull that particular commit from there into your PR.

richm:
> > > From the failing CI, the Fedora is fixed with top commit in #267, debian buster is probably just broken (can be ignored), but centos6 is something that should be fixed as part of this PR as the `man-db` requirement was introduced/changed here.
> > 
> > 
> > ok - do you want to merge #267 then I can rebase?
> 
> Whatever works for you. The #267 changes a lot of things, but they are not used by default so they should not be breaking (but some careful review would be needed). It does not have any deadlines as far as I know.
> 
> If the ostree support needs to land at some reasonable time, feel free to pull that particular commit from there into your PR.

ok - I just fixed the el6 issue - please review - then we can wait on #267 - I did not include that one in my PR - should keep them separate

Jakuje (approved):

PR #271, opened 2023-11-30
docs(changelog): version v0.23.0 [citest skip]

richm:
Update changelog and .README.html for version v0.23.0

### DIFF
--- a/.README.html
+++ b/.README.html
@@ -241,6 +241,7 @@ id="toc-additional-configuration">Additional configuration</a></li>
 Playbook</a></li>
 <li><a href="#template-generation" id="toc-template-generation">Template
 Generation</a></li>
+<li><a href="#rpm-ostree" id="toc-rpm-ostree">rpm-ostree</a></li>
 <li><a href="#license" id="toc-license">License</a></li>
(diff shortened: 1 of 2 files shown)


.README.html
```diff
@@ -664,6 +668,8 @@ <h1 id="template-generation">Template Generation</h1>
 <code>options_match</code>.</p>
 <p>To regenerate the templates, from within the <code>meta/</code>
 directory run: <code>./make_option_lists</code></p>
+<h1 id="rpm-ostree">rpm-ostree</h1>
+<p>See README-ostree.md</p>
```
Jakuje, line 672: references like this are becoming less useful in the generated HTML files. Should we generate also HTML files from the other README-*.md files?

richm: > references like this are becoming less useful in the generated HTML files. Should we generate also HTML files from the other README-*.md files?

maybe - in a subsequent PR

Jakuje (approved):

.README.html
Jakuje: sure. No problem. This was just where I noticed this and given that these changes are landing across several roles, it might be something to consider in the long term.

PR #272, opened 2024-01-08
ci: Use supported ansible-lint action; run ansible-lint against the collection

richm:
The old ansible-community ansible-lint is deprecated.  There is a
new ansible-lint github action.

The latest Ansible repo gating tests run ansible-lint against
the collection format instead of against individual roles.
We have to convert the role to collection format before running
ansible-test.

This also requires tox-lsr 3.2.1

Role developers can run this locally using
`tox -e collection,ansible-lint-collection`
See https://github.com/linux-system-roles/tox-lsr/pull/125

Fix ansible-lint and ansible-test issues reported by the
latest 2.16 versions.

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,6 +1,11 @@
+---
 exclude_paths:
   - tests/roles/
   - .tox/
   - .markdownlint.yaml
 skip_list:
   - var-naming[no-role-prefix]
+mock_roles:
+  - willshersystems.sshd.ansible-sshd
+mock_modules:
+  - ansible.posix.mount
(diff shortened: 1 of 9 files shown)


Jakuje (approved):
Looks good.
The only issue is in the fedora run, which is caused by different formatting of the manual pages in recent fedora. The fix is already in 
https://github.com/willshersystems/ansible-sshd/pull/267/files#diff-be5ce908d6eff907335bb9002a9219ea9c6f57937a36a0bbe7c9448719a258d5
but I can pull it into separate PR if that one is not yet ready.
The debian is likely failing because of missing requirements installation from `tests/requirements.yml`. I can probably do it in separate PR too.

PR #276, opened 2024-02-06
fix: Fix service files generated on EL7 and workaround the tests for containers

Jakuje:
Enhancement: The service files on EL7 are generated with original content.

Reason: The centos7 containers report `ansible_facts["service_mgr"] =  "sysvinit"` which lead to skipping this test in the upstream CI. The CentOS7 is also the only OS that defines the environment file as a mandatory (missing the `-` sign in front of the filename), which caused the downstream tests executed against VMs to fail. But only in the main `sshd.service`. The minus sign is present in the instantiated `sshd@.service` to add even more confusion

Result: The role should generate original service files.

The diff in the generated `sshd.service` file before this change should be limited to the following change:
```diff
-EnvironmentFile=/etc/sysconfig/sshd
+EnvironmentFile=-/etc/sysconfig/sshd
```

### DIFF
--- a/tasks/install_service.yml
+++ b/tasks/install_service.yml
@@ -1,6 +1,9 @@
 ---
 - name: Install systemd service files
-  when: sshd_install_service | bool and ansible_facts['service_mgr'] == 'systemd'
+  when:
+    - sshd_install_service | bool
+    - ansible_facts['service_mgr'] == 'systemd' or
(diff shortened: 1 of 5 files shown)


richm (approved):
Must be a problem with the centos-7 container image - qcow2/qemu works fine

Issue #280, opened 2024-03-18
Setting up second sshd service is broken since last release


Jakuje:
The generated service file will always have the path to the `sshd_main_config_file`, which is defined since 6bb0d7b456eececcf5b5e0c817b113cfde98d170, but was never documented and its usage is somewhat confusing, which I missed initially.

The `sshd_main_config_file` can be overridden by the user, but it is not what was needed to set up the second service before f6ae2094fe23eed94ce64eee1d0f5f1109dfe1da, which used the `sshd_config_file` directly.

The 6bb0d7b456eececcf5b5e0c817b113cfde98d170 makes some operation depending on the following condition which I do not completely understand:
 * `sshd_config_file | dirname != sshd_main_config_file | dirname`

By default the `sshd_config_file` is `/etc/ssh/sshd_config.d/something` and `sshd_main_config_file` is `/etc/ssh/sshd_config`, which means this condition is not matched for default configuration, which is expected. It is used in the following contexts:
 * prevention to write default in the configuration file in drop-in directory
 * creation of drop-in directory
 * checking the main configuration file contains the include directive

It is basically just checking the files are not in the same directory, which works ok if want to detect the second service's configuration file such as `/etc/ssh/sshd_config2`, but fails in cases when the file is in completely different directory, such as `/etc/ssh2/sshd_config`.

Previously, there was a `__sshd_drop_in_dir`, which allowed us to check if the configuration file we are writing is in the drop-in directory to do the above, or its file for separate service in different path, but this does not work anymore

~My proposal would be to refine the above condition to check the `sshd_config_file` is under the dirname of `__sshd_defaults.Include` (it should be able to make up the path even if it is with the star expansion).~

Edit: The first proposal wont work as the `Inlcude` directive is now based on the `sshd_config_file`'s `dirname`. 

More info: https://issues.redhat.com/browse/RHEL-29309

Jakuje:
One option might be to require the `sshd_main_config_file`, which I am not much happy about as this worked before.

Other option might be applying some heuristics to find out if the `sshd_config_file` points to a file which is and which should be included in the main configuration file or not.

I think that the current behavior will even "break" the main sshd service as it will likely try to modify the original `sshd_config` to contain `Include /etc/ssh2/*.conf` if I see right.

Adding the playbook, which is used in the test:
```
---
- name: Configure second SSHD service on different port
  hosts: all
  vars:
    sshd_service: sshd2
[...]

PR #281, opened 2024-03-18
fix: Document and streamline the sshd_main_config_file

Jakuje:
The option was introduced in f6ae2094fe23eed94ce64eee1d0f5f1109dfe1da without documentation and intended use. The recent change f6ae2094fe23eed94ce64eee1d0f5f1109dfe1da propagated this option to the generated service files, which is resulting in unexpected results, when a user decided to set only `sshd_config_file` for the second sshd service such as:
 * the service file points to the system-wide configuration file
 * the system-wide configuration file will get include directive pointing to the directory of the second configuration file

This is an attempt to fix this by introducing some heuristics to guess if the user wants to set up second drop-in directory (ending with .d) or create a standalone configuration file.

Issue Tracker Tickets (Jira or BZ if any): https://issues.redhat.com/browse/RHEL-29309 #280

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,19 @@ When this path points to a drop-in directory (like
 with the variable `sshd_main_config_file`) is checked to contain a proper
 `Include` directive.
 
+#### sshd_main_config_file
+
+When the system is using drop-in directory, this option can be used to set
+a path to the main configuration file and let you configure only the drop-in
+configuration file using `sshd_config_file`. This is useful in cases when
+you need to configure second independent sshd service with different
+configuration file. This is also the file used in the service file.
+
+On systems without drop-in directory, it defaults to `None`. Otherwise it
+defaults to `/etc/ssh/sshd_config`. When the `sshd_config_file` is set
+outside of the drop in directory (its parent directory is not
+`sshd_main_config_file` ~ '.d'), this variable is ignored.
+
 #### sshd_config_namespace
 
 By default (*null*), the role defines whole content of the configuration file
--- a/tests/tests_second_service_drop_in.yml
+++ b/tests/tests_second_service_drop_in.yml
@@ -1,0 +1,122 @@
+---
+- name: Test second sshd service with drop-in directory
+  hosts: all
+  vars:
+    __sshd_test_backup_files:
+      - /etc/ssh/sshd_config
+      - /etc/ssh/sshd_config.d/00-ansible_system_role.conf
+      - /etc/ssh2/sshd_config
+      - /etc/sshd/sshd_config.d/04-ansible.conf
+      - /etc/systemd/system/sshd2.service
+      - /etc/systemd/system/sshd2@.service
+      - /etc/systemd/system/sshd2.socket
+  tasks:
+    - name: "Backup configuration files"
+      ansible.builtin.include_tasks: tasks/backup.yml
+
+    - name: Run the test
+      when:
+        - (ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version'] | int > 8) or
+          (ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_major_version'] | int >= 20)
+      block:
+        - name: Create ssh2 directory
+          ansible.builtin.file:
+            path: /etc/ssh2
+            state: directory
+            mode: '0755'
+
+        - name: Create sshd_config file
+          ansible.builtin.file:
+            path: /etc/ssh2/sshd_config
+            state: touch
+            mode: '0600'
+
+        - name: Configure alternative sshd_config file
+          ansible.builtin.include_role:
+            name: ansible-sshd
+          vars:
+            sshd_service: sshd2
+            sshd_main_config_file: /etc/ssh2/sshd_config
+            sshd_config_file: /etc/ssh2/sshd_config.d/04-ansible.conf
+            sshd_install_service: true
+            sshd_manage_selinux: true
+            sshd:
+              Port: 2222
+              ForceCommand: echo "CONNECTED2"
+
+        - name: Verify the config options are correctly set
+          tags: tests::verify
+          block:
+            - name: Flush handlers
+              ansible.builtin.meta: flush_handlers
+
+            - name: Stat the parent directory
+              ansible.builtin.stat:
+                path: /etc/ssh2
+              register: parent_stat
+
+            - name: Print the main configuration file
+              ansible.builtin.slurp:
+                src: /etc/ssh2/sshd_config
+              register: config
+
+            - name: Print the drop-in configuration file
+              ansible.builtin.slurp:
+                src: /etc/ssh2/sshd_config.d/04-ansible.conf
+              register: config_drop_in
+
+            - name: Check content of the created configuration file
+              ansible.builtin.assert:
+                that:
+                  - "'Port 2222' in config_drop_in.content | b64decode"
+                  - "'ForceCommand echo' in config_drop_in.content | b64decode"
+
+            - name: Check Include is present in the main configuration file
+              ansible.builtin.assert:
+                that:
+                  - "'Include' in config.content | b64decode"
+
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -209,6 +209,17 @@ When this path points to a drop-in directory (like
...
+#### sshd_main_config_file
+
+When the system is using drop-in directory, this option can be used to set
+a path to the main configuration file and let you configure only the drop-in
+configuration file using `sshd_config_file`. This is useful in cases when
+you need to configure second sshd service with different configuration file.
+This is also the file used in the service file.
+
+On systems with drop-in directory, it defaults to `/etc/ssh/sshd_config`,
+otherwise to `None`.
```
richm, line 221: Do we need to add a note like "This assumes that the name of the drop-in directory is the same as the name of the config file with '.d' appended.  For example, if the main config file is is `/etc/ssh/sshd_config`, this assumes the drop-in directory is `/etc/ssh/sshd_config.d`."?

Jakuje: Likely we should add something about that. I wrote it before writing the fix ...

This is the most common drop-in directory naming I am familiar with, but it should be mentioned.

Jakuje: Attempted to clarify this in the updated changeset

richm (approved):

nkakouros:
I think I understand the issue now. The problem is that for a distro, the role uses a default `sshd_main_config_file`, but the user may not want to use that. The user has to explicitly set `sshd_main_config_file` to `None` to prevent the new sshd service from using the OS's default. Is this right?

In this case, I wonder if the role could do:

```yaml
- set_fact:
    sshd_main_config_file: None
  when: sshd_config_file == sshd_main_config_file ~ '.d'
```

and the rest of the code stays as is already.

Jakuje:
Thank you for having a look into this!

> ```yaml
> - set_fact:
>     sshd_main_config_file: None
>   when: sshd_config_file | dirname == sshd_main_config_file ~ '.d'
> ```

Yes, this could be the other option, maybe nicer as it would be checked only in one place, unlike what I suggested.

But still I am not sure about the condition `sshd_config_file | dirname != sshd_main_config_file | dirname`. I would probably keep it in the way I suggested as `sshd_config_file | dirname == sshd_main_config_file ~ '.d'`, what do you think?

I will try to put together some tests coverage for the change here too.

Jakuje:
> But still I am not sure about the condition `sshd_config_file | dirname != sshd_main_config_file | dirname`. I would probably keep it in the way I suggested as `sshd_config_file | dirname == sshd_main_config_file ~ '.d'`, what do you think?

My bad. All these exist in condition with `and sshd_main_config_file is not none` so they could be removed altogether if I see right when we will make sure the `sshd_main_config_file` is `None`.

Jakuje:
The problem with the `set_fact` is that the set fact and variable is propagated to subsequent role invocation, for example if we run the tests like we do now with single `ansible-playbook`, the `set_fact` in first invocation is kept around for next test, and later, causing unexpected failures.

We had issues with others variables that were force set as facts in the past. I think the solution was to set only internal variables or to avoid `set_fact` altogether.

So unless there will be some better ideas, I will revert the changes to the previous version, even though it requires modification on more places. Testing and feedback welcomed.

tests/tests_second_service_drop_in.yml
```diff
@@ -0,0 +1,137 @@
...
+      ansible.builtin.include_tasks: tasks/backup.yml
+
+    - name: Create ssh2 directory
+      ansible.builtin.file:
+        path: /etc/ssh2
+        state: directory
+        mode: '0755'
+      when:
+        - (ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version'] | int > 8) or
+          (ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_major_version'] | int >= 20)
```
richm, line 24: To avoid having to put this `when` condition on every task - you could do
```yaml
  tasks:
    - name: This test is only supported on later versions of RedHat family and Ubuntu
      meta: end_host
      when:
        - not ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version'] | int > 8
        - not ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_major_version'] | int >= 20
```

richm: hmm - maybe not - I see you have a different condition below

tests/tests_second_service_drop_in.yml
```diff
@@ -0,0 +1,137 @@
...
+              - parent_stat.stat.exists
+              - parent_stat.stat.mode == '0755'
+
+    - name: Verify the service files are correct
+      tags: tests::verify
+      when:
+        - ansible_facts['service_mgr'] == 'systemd' or
+          (ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version'] == '7')
+        - (ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version'] | int > 8) or
+          (ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_major_version'] | int >= 20)
```
richm, line 98: If `ansible_facts['os_family'] == 'RedHat'` and `ansible_facts['distribution_major_version'] == '7'` then I don't think this block will ever run

nkakouros:
> The problem with the `set_fact` is that the set fact and variable is propagated to subsequent role invocation, for example if we run the tests like we do now with single `ansible-playbook`, the `set_fact` in first invocation is kept around for next test, and later, causing unexpected failures.

To work around this you can do:

```
# as the last task of the role
- set_fact:
    sshd_main_config_file: __sshd_sentinel__

# in variables.yml
- set_fact:
    sshd_main_config_file: "{{ __sshd_main_config_file }}"
  when: sshd_main_config_file == '__sshd_sentinel__'

- set_fact: 
    sshd_main_config_file: None
  when: sshd_config_file | dirname == sshd_main_config_file ~ '.d'
```

I am not saying this is less cognitive load for a person that reads the role's code, but it keeps things centralized.

tests/tests_second_service_drop_in.yml
Jakuje: The problem with `end_host` was that it was not supported by RHEL6 if I remember well.

tests/tests_second_service_drop_in.yml
Jakuje: Good point. Will remove that part.

tests/tests_second_service_drop_in.yml
Jakuje: I can probably wrap it in `block` to avoid the repetition.

Jakuje:
> To work around this you can do:
> 
> ```yaml
> # as the last task of the role
> - set_fact:
>     sshd_main_config_file: __sshd_sentinel__
> 
> # in variables.yml
> - set_fact:
>     sshd_main_config_file: "{{ __sshd_main_config_file }}"
>   when: sshd_main_config_file == '__sshd_sentinel__'
> 
> - set_fact: 
>     sshd_main_config_file: None
>   when: sshd_config_file | dirname == sshd_main_config_file ~ '.d'
> ```
> 
> I am not saying this is less cognitive load for a person that reads the role's code, but it keeps things centralized.

I am still undecided if having fact set to placeholder variable is nicer or more fragile. @richm @mattwillsher any thoughts or strong opinions?

richm:
In general it is a bad idea for the role runtime code to do

```yaml
set_fact:
  sshd_role_variable: value
```

for role variables declared in `defaults/main.yml` - it changes the scope from whatever was the original scope to global, and may cause weird, unexpected behavior if the role is used more than once in a play.

richm:
any update?

Jakuje:
I think this is good to go if there are no other concerns

richm (approved):

mattwillsher (approved):

Jakuje:
Thank you for the reviews and comments. Merging then.

PR #290, opened 2024-06-03
feat:  Ubuntu noble

mattwillsher:
Enhancement: 

* Added Ubuntu 24.04/Noble support
* Removed CentOS 6 & 8 tests due to them being EOL
* Fix a bunch of test issues

Reason:

Result:


mattwillsher:
Hi @Jakuje 

Current failure is down to the sockets systemd unit file. Ubuntu 24.04 default is:

```ini
[Unit]
Description=OpenBSD Secure Shell server socket
Before=sockets.target ssh.service
ConditionPathExists=!/etc/ssh/sshd_not_to_be_run

[Socket]
ListenStream=22
Accept=no
FreeBind=yes

[Install]
WantedBy=sockets.target
RequiredBy=ssh.service
```

Our template is 
```
[Unit]
Description=OpenBSD Secure Shell server socket
Documentation=man:sshd(8) man:sshd_config(5)
{% if __sshd_socket_accept %}
Conflicts={{ sshd_service }}.service
{% else %}
Before=sockets.target
{% endif %}

[Socket]
ListenStream=22
{% if __sshd_socket_accept %}
Accept=yes
{% else %}
Accept=no
{% endif %}

[Install]
WantedBy=sockets.target
```

Options I can see are:
* Add conditions for Ubuntu 24.04 to the template
* Convert code to clone vendor provided sshd.socket and use lineinfile to replace the ones we want to control
* Generate a systemd override file rather than shadowing the entire file - I've only had a cursory look at this, so maybe this isn't a viable option?

What are your thoughts?

(P.S. No more PRs. the misspelling of noble on the branch was annoying me and source branch can't be changed on a PR)

Jakuje:
> Hi @Jakuje
> 
> Current failure is down to the sockets systemd unit file. Ubuntu 24.04 default is:

Thank you for having a look into this!

> ```ini
> [Unit]
> Description=OpenBSD Secure Shell server socket
> Before=sockets.target ssh.service
> ConditionPathExists=!/etc/ssh/sshd_not_to_be_run
> 
> [Socket]
> ListenStream=22
> Accept=no
> FreeBind=yes
> 
> [Install]
> WantedBy=sockets.target
> RequiredBy=ssh.service
> ```

I am wondering how it even works ... Its already some time since I was maintaining the OpenSSH, but this sounds like both ssh.service and ssh.socket is enabled which in the past caused random failures (whether systemd or openssh got hold of the port 22 first). But this seems to be solved by ordering the socket first with `RequiredBy` and also the `FreeBind` which will allow binding to the addresses that are not up yet ...

> Our template is
> 
> ```
> [Unit]
> Description=OpenBSD Secure Shell server socket
> Documentation=man:sshd(8) man:sshd_config(5)
> {% if __sshd_socket_accept %}
> Conflicts={{ sshd_service }}.service
> {% else %}
> Before=sockets.target
> {% endif %}
> 
> [Socket]
> ListenStream=22
> {% if __sshd_socket_accept %}
> Accept=yes
> {% else %}
> Accept=no
> {% endif %}
> 
> [Install]
> WantedBy=sockets.target
> ```

> Options I can see are:
> 
>     * Add conditions for Ubuntu 24.04 to the template

I think that would be preferred. If I see right, we need `ConditionPathExists`, `RequiredBy`, `FreeBind` as well as adjusting the `Before`.

>     * Convert code to clone vendor provided sshd.socket and use lineinfile to replace the ones we want to control

This could do as we are writing the files in `/etc/` and the systemd has the package-installed service files in `/usr/lib/systemd/system/ssh.service`. But we would have to depend on these being intact on the target system, which is something I wanted to avoid by keeping the results independent on the potentially/partially broken systems.

>     * Generate a systemd override file rather than shadowing the entire file - I've only had a cursory look at this, so maybe this isn't a viable option?

The override works ok when we want to modify some features in existing service, but if we want to set up separate `sshd` service with different name and different configuration, we still need to generate the full unit file anyway (I think). So we would need to have to implement both options anyway.

Jakuje:
Regarding the CentOS 8, I got https://github.com/roles-ansible/check-ansible-centos-centos8-action/pull/7 merged so it should keep working with the archive repositories. Certainly not for production use, but it should be good enough for CI use.

Your PR for the ubuntu was also merged to the actions so you can revert that / drop that patch too when you will be doing next changes.

README.md
```diff
@@ -27,7 +27,7 @@ via simple password. If you need this functionality, be sure to set
 
 Tested on:
 
-* Ubuntu precise, trusty, xenial, bionic, focal, jammy
+* Ubuntu precise, trusty, xenial, bionic, focal, jammy, nobel
```
Jakuje, line 30: typo:
```suggestion
* Ubuntu precise, trusty, xenial, bionic, focal, jammy, noble
```

meta/main.yml
```diff
@@ -18,14 +18,15 @@ galaxy_info:
...
-        - precise
-        - trusty
-        - xenial
         - bionic
         - focal
         - jammy
+        - noble
+        - precise
+        - trusty
+        - xenial
```
Jakuje, line 27: I would keep them in the verion order instead of the alphabetical as the letters wrapped around between xenial (16.04) and bionic (18.04). Moving these on the bottom after noble (24.04) will just make it more confusing in my opinion (especially when we will add the next lts version starting on a letter `R` somewhere in the middle.

templates/sshd.socket.j2
```diff
@@ -14,6 +14,12 @@ Accept=yes
...
 {% endif %}
+{% if ansible_facts["distribution"]=="Ubuntu" and ansible_facts["distribution_version"] is version('24.04', '>=') %}
+FreeBind=yes
+{% endif %}
 
 [Install]
 WantedBy=sockets.target
+{% if ansible_facts["distribution"]=="Ubuntu" and ansible_facts["distribution_version"] is version('24.04', '>=') %}
+RequiredBy=ssh.service
+{% endif %}
```
Jakuje, line 25: I would prefer to add new internal variables such as `__sshd_socket_freebind` and `__sshd_socket_required_by` (naming can be improved if you can come up with nicer ones) to keep this version information contained in the vars files instead of here. Its just matter of time this will land in Debian and when some other distributions will pick something similar or same.

tests/tests_systemd_services.yml
```diff
@@ -94,6 +94,7 @@
         #  * I do not think the ConditionPathExists is much useful so skipping on Ubuntu
         #  * Before= does not make any sense in combination with Conflicts=
         #  * I do not think the Description needs to match verbatim either
+        #  * Accept varies in the default file between Ubuntu versions and is explicit anyway
```
Jakuje, line 97: Why do you exclude the Accept here? I thought the Accept was already handled. In any case, its value makes quite a huge difference.

mattwillsher:
@Jakuje I think this is about there. I've disable the tests for the ssh@.service as 24.04 doesn't have the file in it's package. Please review.

tests/tests_systemd_services.yml
```diff
@@ -118,6 +120,7 @@
         - ansible_facts['service_mgr'] == 'systemd' or
           (ansible_facts['os_family'] == 'RedHat' and ansible_facts['distribution_major_version'] == '7')
         - ansible_facts['distribution'] != "Debian" or ansible_facts['distribution_major_version'] | int < 12
+        - not (ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_version'] is version('24.04', '>=')) # 24.04 has no sshd@.server in the package
```
Jakuje, line 123:
```suggestion
        - not (ansible_facts['distribution'] == 'Ubuntu' and ansible_facts['distribution_version'] is version('24.04', '>=')) # 24.04 has no sshd@.socket in the package
```

mattwillsher: @.server is correct, see code below the comment.

vars/Ubuntu_24.yml
```diff
@@ -28,5 +28,5 @@ __sshd_environment_variable: $SSHD_OPTS
 __sshd_service_after: auditd.service
 __sshd_service_alias: sshd
 
-__sshd_socket_freebind: "yes"
+__sshd_socket_freebind: 'yes'
```
spetrosi: Can it be a boolean `true`?

mattwillsher: It is on newer commits. Will fix the test failures in the coming day or two

tests/tests_systemd_services.yml
mattwillsher: I missed the value from the vars. Thats now set, this change has been reverted.

Jakuje (approved):
lgtm!

PR #293, opened 2024-07-02
fix: add support for EL10

richm:
According to the Ansible team, support for listing platforms in
role `meta/main.yml` files is being removed.
Instead, they recommend using `galaxy_tags`

https://github.com/ansible/ansible/blob/stable-2.17/changelogs/CHANGELOG-v2.17.rst
"Remove the galaxy_info field platforms from the role templates"
https://github.com/ansible/ansible/issues/82453

For each version listed under `platforms.EL` - add a tag like `elN`.

Q: Why not use a delimiter between the platform and the version e.g. `el-10`?

This is not allowed by ansible-lint:

```
meta-no-tags: Tags must contain lowercase letters and digits only., invalid: 'el-10'
meta/main.yml:1
```

So we cannot use uppercase letters either.

Q: Why not use our own meta/main.yml field?

No other fields are allowed by ansible-lint:

```
syntax-check[specific]: 'myfield' is not a valid attribute for a RoleMetadata
```

Q: Why not use some other field?

There are no other applicable or suitable fields.

Q: What happens when we want to support versions like `N.M`?

Use the word "dot" instead of "." e.g. `el10dot3`.
Similarly - use "dash" instead of "-".

We do not need tags such as `fedoraall`.
The `fedora` tag implies that the role works on all supported versions of fedora.
Otherwise, use tags such as `fedora40` if the role only supports specific versions.

In addition - for roles that have different variable files for EL9, create
the corresponding EL10 files, and update the variables for EL10.

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -35,4 +35,4 @@ jobs:
       - name: Run ansible-lint
         uses: ansible/ansible-lint@v24
         with:
-          working_directory: .tox/ansible_collections/${{ env.LSR_ROLE2COLL_NAMESPACE }}/${{ env.LSR_ROLE2COLL_NAME }}
+          working_directory: ${{ github.workspace }}/.tox/ansible_collections/${{ env.LSR_ROLE2COLL_NAMESPACE }}/${{ env.LSR_ROLE2COLL_NAME }}
--- a/.ostree/packages-runtime-CentOS-10.txt
+++ b/.ostree/packages-runtime-CentOS-10.txt
@@ -1,0 +1,1 @@
+packages-runtime-RedHat-10.txt
--- a/.ostree/packages-runtime-CentOS-6.txt
+++ b/.ostree/packages-runtime-CentOS-6.txt
@@ -1,2 +1,1 @@
-openssh
-openssh-server
+packages-runtime-RedHat-6.txt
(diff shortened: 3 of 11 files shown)


richm:
Note - I wanted to add an ansible check github action for centos10 - there isn't one - the centos-latest check still points to quay.io centos:latest which is centos 8 (!)

Jakuje:
Please, update also readme with the el10.

> Note - I wanted to add an ansible check github action for centos10 - there isn't one

I would find useful to have something that follows `stream10-development` tag, but right now its mostly whats in Fedora so I am ok with the change without the centos10 check for now. You will run something downstream anyway.

> the centos-latest check still points to quay.io centos:latest which is centos 8 (!)

Well ... thats the latest centos released ... others are streams: https://quay.io/repository/centos/centos?tab=tags&tag=latest

Regarding the failing CentOS7, its EOL so the mirrors were moved. If we would like to run tests against it, we will have to fix the mirrors similar way as for CentOS6 and CentOS8. Can you check if this works?

https://github.com/roles-ansible/check-ansible-centos-centos7-action/pull/2

Jakuje (approved):
Thanks. Lets wait some time if the change to the action get merged. Given my ack for now.

richm:
> Thanks. Lets wait some time if the change to the action get merged. Given my ack for now.

I can revert https://github.com/willshersystems/ansible-sshd/pull/293/commits/9a7c358c8583e7343c0d96e2a76580b993b54d4b from the PR, then we can merge.

PR #298, opened 2024-09-06
fix: use quote with command, shell and validate with variable

Jakuje:
This is the replacement for the stale PR #245, where the reporter never came back to fix the CI issues, but we still consider that useful change.

Enhancement: Use quote command when calling shell/command modules.

Reason: The user suplied variables could split if they contain whitespace and cause unexpected behavior.

Result: User input is validated.

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -80,7 +80,7 @@
         {% if sshd_sysconfig %}
           source /etc/sysconfig/sshd
         {% endif %}
-        ssh-keygen -q -t {{ item | regex_search('(rsa|dsa|ecdsa|ed25519)') }} -f {{ item }} -C '' -N ''
+        ssh-keygen -q -t {{ item | regex_search('(rsa|dsa|ecdsa|ed25519)') }} -f {{ item | quote }} -C '' -N ''
       args:
         creates: "{{ item }}"
       loop: "{{ __sshd_verify_hostkeys | from_json | list }}"
@@ -109,7 +109,7 @@
 
     - name: Generate temporary hostkey
       ansible.builtin.command: >
-        ssh-keygen -q -t rsa -f '{{ sshd_test_hostkey.path }}/rsa_key' -C '' -N ''
+        ssh-keygen -q -t rsa -f {{ sshd_test_hostkey.path | quote }}/rsa_key -C '' -N ''
       changed_when: false
       when: sshd_test_hostkey.path is defined
 
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -4,22 +4,8 @@
   when:
     - not __sshd_os_supported | bool
 
-- name: Ensure variables used in templates in shell or systemd unit contexts are okay
-  ansible.builtin.assert:
-    that:
-      - (sshd_sysconfig_use_strong_rng | string) == (sshd_sysconfig_use_strong_rng | quote)
-      - sshd_binary == (sshd_binary | quote)
-      - sshd_config_file == (sshd_config_file | quote)
-      - (__sshd_runtime_directory is none) or (__sshd_runtime_directory == (__sshd_runtime_directory | quote))
-      - __sshd_runtime_directory_mode == (__sshd_runtime_directory_mode | quote)
-    msg: |
-      sshd_sysconfig_use_strong_rng: {{ sshd_sysconfig_use_strong_rng }} == {{ sshd_sysconfig_use_strong_rng | quote }}
-      sshd_binary: {{ sshd_binary }} == {{ sshd_binary | quote }}
-      sshd_config_file: {{ sshd_config_file }} == {{ sshd_config_file | quote }}
-      {%- if __sshd_runtime_directory is not none %}
-      __sshd_runtime_directory: {{ __sshd_runtime_directory }} == {{ __sshd_runtime_directory | quote }}
-      {% endif %}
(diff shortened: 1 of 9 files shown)


tasks/install.yml
```diff
@@ -4,6 +4,23 @@
...
 
+- name: Ensure variables used in templates in shell or systemd unit contexts are okay
+  ansible.builtin.assert:
+    that:
+      - (sshd_sysconfig_use_strong_rng | string) == (sshd_sysconfig_use_strong_rng | quote)
+      - sshd_binary == (sshd_binary | quote)
+      - sshd_config_file == (sshd_config_file | quote)
+      - (__sshd_runtime_directory is none) or (__sshd_runtime_directory == (__sshd_runtime_directory | quote))
+      - __sshd_runtime_directory_mode == (__sshd_runtime_directory_mode | quote)
+    msg: |
```
spetrosi, line 15: It would be nice to document what users must do in the case this task fails here in the error message.

tests/tests_hostkeys_unsafe_path.yml
```diff
@@ -0,0 +1,68 @@
...
+        state: directory
+        path: "{{ ansible_facts.env.TMPDIR }}"
+        mode: '0755'
+
+    - name: Configure sshd with BAD config
+      ansible.builtin.include_role:
+        name: ansible-sshd
+      vars:
+        sshd_skip_defaults: true
+        sshd_verify_hostkeys: []
```
Jakuje, line 44: Seems like the ansible (2.16) in CentOS8 has an issue with this and fails like
```
TASK [ansible-sshd : Install ssh packages] *************************************
fatal: [localhost]: FAILED! => {"changed": false, "module_stderr": "  File \"<stdin>\", line 6\n    modlib_path = '/tmp/a $(touch -- \"$(echo L3RtcC9CQURGTEFH | base64 -d)\") ' $(touch -- \"$(echo L3RtcC9CQURGTEFH | base64 -d)\") '\" $(touch -- \"$(echo L3RtcC9CQURGTEFH | base64 -d)\") \"b/ansible_ansible.legacy.dnf_payload_3lf_209j/ansible_ansible.legacy.dnf_payload.zip'\n                                                                               ^\nSyntaxError: invalid syntax\n", "module_stdout": "", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 1}
```
Given that the CentOS 8 is EOL, I do not think it makes sense to fill a bug for Ansible so I think we should just skip running this test on EL8, what do you think?

tasks/install.yml
Jakuje: Thanks for the suggestion! Indeed, this would need some more description and likely a negative test. Will look into that!

tests/tests_hostkeys_unsafe_path.yml
spetrosi: CentOS 8 (and 7) need a workaround to use a vault repository, like in https://github.com/spetrosi/tft-tests/blob/main/plans/general.fmf#L33-L40

Jakuje: That is already done in the GH action code, but this error is most likely in the ansible itself. All the other tests work on centos8stream CI just ok.

richm: Is the problem here that the `TMPDIR` env. var. is defined for *all* modules, including those that we don't want to test?

Jakuje: Probably yes. I do not know how to check this nor how this can be changed. See the error log in the first comment of this thread for more information.

Jakuje: @richm not sure if your comment suggests "this is ok" or "this test is bad and we should remove it" or something in between. I would like to avoid setting some testing stuff in `tasks/*.yml` though.

richm: > @richm not sure if your comment suggests "this is ok" or "this test is bad and we should remove it" or something in between.

I think something in between.  The test as it is doesn't work because it breaks `package` and `service` tasks, so we can't use it.

> I would like to avoid setting some testing stuff in `tasks/*.yml` though.

Me too, but I don't know of a way to set env `TMPDIR` for every task other than `package` and `service`.  Maybe module defaults?  https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_module_defaults.html - although that mechanism can be used to set per-task `environment` - looks like just the arguments to the actual module being used.

Jakuje: >  The test as it is doesn't work because it breaks package and service tasks, so we can't use it.

But only for RHEL8 -- on others it seems to be handled ok if I see right. Thats why I skipped it there:

https://github.com/willshersystems/ansible-sshd/pull/298/files#diff-937ad845535afbcf53affe09def9df8c3d3cf47bbadd5bdddf47c5f4443bb33eR46

richm: ok

richm (approved):

mattwillsher (approved):
Thanks @Jakuje !

mattwillsher:
Is this ready for merge?

Jakuje:
Yes, I do not have anything else to add here. Feel free to merge when needed.

Issue #297, opened 2024-09-04
ansible-lint complains about sshd var naming


bayramkzk:
I am using `ansible-lint` in Neovim and linter complains about `sshd` variable naming convention here:

```yaml
    - name: Configure sshd server daemon config
      ansible.builtin.include_role:
        name: willshersystems.sshd
      vars:
        sshd: # here
          PubkeyAuthentication: true
          PasswordAuthentication: false
          ClientAliveInterval: 60
          Port: "{{ my_sshd_port }}"
        sshd_manage_firewall: true
```

```
site.yaml|27 col 1 error| Variables names from within roles should use sshd_ as a prefix. (vars: sshd)
```

Linter doesn't complain when, say, I rename `sshd` to `sshd_config` but as expected the playbook doesn't work properly then.

mattwillsher:
Hi @bayramkzk 
The [.ansible-lint file excludes this check](https://github.com/willshersystems/ansible-sshd/blob/de66f6065994e79f8caa744766052d503b67c3cb/.ansible-lint#L7). 
There is issue #246 with a view to rename the variables, but that may not be implemented due to the extent of changes required and risks to existing use of the role.

mattwillsher:
I'm going to reopen this as it's different to the issue seen in #246 as this is linting from the calling play rather than the module itself. I can't give time scales on the fix, but [@spetrosi's suggestion ](https://github.com/willshersystems/ansible-sshd/issues/246#issuecomment-2329514112) looks like a good way forward.

Jakuje:
Ugh. It looks like the `sshd` variable itself is an issue for ansible. Not sure how we should name it then. I can think of `sshd_` but I feel that would be just a workaround for this policy and would make it just more confusing for everyone.

The obvious workaround is:
```
      vars:
        sshd_PubkeyAuthentication: true
        sshd_PasswordAuthentication: false
        sshd_ClientAliveInterval: 60
        sshd_Port: "{{ my_sshd_port }}"
```
but I like the first syntax much more.

mattwillsher:
`sshd_config` looks like it's not used. So we could take [this suggestion](https://github.com/willshersystems/ansible-sshd/issues/246#issuecomment-2329514112) to deprecate `sshd`, update the docs etc. to `sshd_config`. 
Or we could use `sshd_sshd`, which looks weird but should stop the linter complaining.

mattwillsher:
I'll pick this up and get a PR put together.

[...]

Issue #246, opened 2023-08-08
chore(linting): update variable names to include sshd_ prefix


mattwillsher:
Var names should be prefixed with role name to pass lint checks.

If no time to work on this check can be disabled for now.

Jakuje:
This was disabled in 34be8e5ea372f25c4ed643340e8b9416eeae0738 if I see right

mattwillsher:
The lint issue I've come across recently on a new role but try as I might I couldn't get the name to comply. 
I do believe that we should conform to this rule, though, as it's effectively a variable scoping challenge. 

I'll leave this open for now as a reminder to myself.

spetrosi:
In system roles, we deprecated old variables by printint a message to users who use those old variables, and then setting the correct variable name, see https://github.com/linux-system-roles/mssql/blob/main/tasks/main.yml#L12-L25
Then, maybe in a couple years, we can remove the deprecated variable completely.

Jakuje:
This was resolved with #299. Closing.

PR #299, opened 2024-09-12
fix: rename var sshd -> sshd_config and debug output

mattwillsher:
Enhancement:

Address role prefix on variable names.

Reason: Coding standards
Result:

 #297,  #246

### DIFF
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -33,8 +33,7 @@
     - __sshd_hostkeys_nofips | d([])
 
 - name: Make sure hostkeys are available and have expected permissions
-  vars:
-    &share_vars # 'MAo=' evaluates to '0\n' in base 64 encoding, which is default
+  vars: &share_vars # 'MAo=' evaluates to '0\n' in base 64 encoding, which is default
     __sshd_fips_mode: >-
       {{ __sshd_hostkeys_nofips | d([]) and
          (__sshd_kernel_fips_mode.content | d('MAo=') | b64decode | trim == '1' or
--- a/tasks/install.yml
+++ b/tasks/install.yml
@@ -33,8 +33,8 @@
     - __sshd_hostkeys_nofips | d([])
 
 - name: Make sure hostkeys are available and have expected permissions
-  vars: &share_vars
-    # 'MAo=' evaluates to '0\n' in base 64 encoding, which is default
+  vars:
+    &share_vars # 'MAo=' evaluates to '0\n' in base 64 encoding, which is default
     __sshd_fips_mode: >-
       {{ __sshd_hostkeys_nofips | d([]) and
          (__sshd_kernel_fips_mode.content | d('MAo=') | b64decode | trim == '1' or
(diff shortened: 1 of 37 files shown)


tasks/install.yml
```diff
@@ -33,8 +33,8 @@
     - __sshd_hostkeys_nofips | d([])
 
 - name: Make sure hostkeys are available and have expected permissions
-  vars: &share_vars
-    # 'MAo=' evaluates to '0\n' in base 64 encoding, which is default
+  vars:
+    &share_vars # 'MAo=' evaluates to '0\n' in base 64 encoding, which is default
```
mattwillsher, line 37: VSCode's YAML extension made this change, so I assume it's a safe noop change.

tasks/main.yml
```diff
@@ -1,4 +1,17 @@
...
+    - name: Print that the sshd variable is deprecated
+      ansible.builtin.debug:
+        msg: >-
+          The sshd variable is deprecated and will be removed
+          in a future version. Edit your playbook to use
+          the sshd_config variable instead.
+
+    - name: Link the deprecated sshd fact
+      ansible.builtin.set_fact:
+        sshd_config: "{{ sshd }}"
```
richm, line 14: I don't see any other way to do this, but I think it is important to note that this will cause `sshd_config` to be set in the global namespace, which may be in a different scope than `sshd`.  For example, if you did something like this:
```yaml
- name: Manage sshd settings
  include_role:
    name: willshersystems.ansible-sshd
  vars:
    sshd:
      setting:....
```
the variable `sshd` is scoped locally to the role invocation, but `sshd_config` will be set in the global namespace.  This may cause some subtle problems.

mattwillsher: Thanks Richard. Something is causing the default tests to fail, I wonder if it's this?
Maybe something like the following in defaults and use `__sshd_config` through the rest of the role?  

```yaml
__sshd_config: "{{ sshd_config | default({}) or sshd | default({}) }}"
```

mattwillsher: The above assignment vars seems to work. Happy days.

tasks/install.yml
mattwillsher: All test still pass with the changed format.

Jakuje (approved):
Did not notice any issue. Maybe just a suggestion to provide some simple sanity test that the old name keeps working (ideally with both `role` and `include_role` syntax`).

richm:
> Did not notice any issue. Maybe just a suggestion to provide some simple sanity test that the old name keeps working (ideally with both `role` and `include_role` syntax`).

+1

tests/tests_deprecated_sshd_variable.yml
```diff
@@ -0,0 +1,113 @@
...
+    - name: Configure sshd
+      ansible.builtin.include_role:
+        name: ansible-sshd
+      vars:
+        sshd:
+          AcceptEnv: LANG
+          Banner: /etc/issue
+          Ciphers: aes256-ctr
+          Subsystem: "sftp internal-sftp"
+        sshd_config_file: /etc/ssh/sshd_config
```
Jakuje, line 21: This line is causing the issues for  alpine and ubuntu latest as when you generate the main sshd_config without hostkeys, the `sshd -T` will not work.

mattwillsher: My assumption the operation of -T has changed in a recent OpenSSH release or within the packages, however if I install and run `sshd -T` in an ubuntu:latest Docker container I get a different error (`Missing privilege separation directory: /run/sshd`) until sshd is started. This is also the case on 24.04. Host key missing occurs when `sshd -T` is run as an unprivileged user, though that isn't the case in the tests. Host keys exist on ubuntu:latest, set up at package install time.

I've not had time to investigate any further, and don't have an ideas of what the fix will be.

Jakuje: Sorry, the new ubuntu image is missing the `unminimize` command by default so it needs to be installed first. Hope it will work afterward.

The case of alpine, it might be something else. I will try to investigate a bit.

Jakuje: I pushed 2 commits that should resolve these two issues. Please, take a look.

mattwillsher:
Thanks @Jakuje, you're a star 🌟

Issue #302, opened 2024-10-30
sshd restart required after crypto policy update


meta-metaferia:
On RHEL8, when these variables are set to `true`
```
sshd_sysconfig: true
sshd_sysconfig_override_crypto_policy: true
```
It goes and update this file `/etc/sysconfig/sshd` . However, at the end of the role execution it only does an sshd reload which is not enough for the change made inside  `/etc/sysconfig/sshd` to take effect, a restart of the sshd service is required.

Jakuje:
Good catch! Sorry for a delay though. I filled #303. Can you check it will solve your issue?

Issue #236, opened 2023-05-07
SSH is reloaded by default even if sshd_manage_service is set to false


Ouack23:
Hello,

The documentation states that :
```markdown
* `sshd_allow_reload`

If set to *false*, a reload of sshd wont happen on change. This can help with
troubleshooting. You'll need to manually reload sshd if you want to apply the
changed configuration. Defaults to the same value as `sshd_manage_service`.
(Except on AIX, where `sshd_manage_service` is default *false*, but
`sshd_allow_reload` is default *true*)
```

but the default variables set `sshd_allow_reload` to `true` : https://github.com/willshersystems/ansible-sshd/blob/master/defaults/main.yml#L14

So either the documentation is wrong and should be corrected, or the defaults file should be changed like this:
```diff
# If the below is false, don't reload the ssh daemon on change
- sshd_allow_reload: true
+ sshd_allow_reload: "{{ sshd_manage_service }}"
```

Thank you for reading :)

Jakuje:
When this was introduced in bcd864fea4ff250716137465f36f7fdc6ee11ce5, the documentation was matching the code. But it is "broken" since 43ed7c19a2be6f183b60dd8a068e809178a194b8 which dropped the assignment.

I am worried that the problem with such assignment might have issues with the variable precedence (would have to double-check though). Do you want to open a PR with this change, ideally also with a reproducer tests to verify that it works as expected in the testing environment (probably only by checking variables as they are based on containers and do not have services running)?

Jakuje:
@mattwillsher the change wont be that easy as proposed in the issue. I think the defaults/main.yml are evaluated before the user variables are considered so it might be easier to update documentation ...

richm:
> @mattwillsher the change wont be that easy as proposed in the issue. I think the defaults/main.yml are evaluated before the user variables are considered so it might be easier to update documentation ...

It depends - variables defined in defaults/main.yml, and in `vars` (which includes `vars/main.yml` as well as inline `vars:`) are lazily evaluated - they are only evaluated when the value is actually needed - and `defaults/main.yml` has the lowest precedence https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#ansible-variable-precedence - so it might work

mattwillsher:
I was thinking update the docs, which seems to be the original ask?

[...]

PR #303, opened 2024-12-16
fix: Reload the service when needed

Jakuje:
The change of sysconfig requires the full restart instead of only reload.

I did not figure out a better way to avoid doing both in this case, so comments and suggestions welcomed.

Fixes: #302

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ Ubuntu. This is not the default assigned by this module - it will set
 via simple password. If you need this functionality, be sure to set
 `sshd_PermitRootLogin yes` for those hosts.
 
+**NOTE** The sshd service is reloaded/restarted automatically, only if the role is
+invoked using `roles` keyword. Using `include_role` won't trigger handlers
+as described in
+[official documentation](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_handlers.html#controlling-when-handlers-run).
+If you need to invoke the handlers in this case, use `meta: flush_handlers`.
+
 ## Requirements
 
 Tested on:
--- a/README.md
+++ b/README.md
@@ -96,6 +96,14 @@ changed configuration. Defaults to the same value as `sshd_manage_service`.
 (Except on AIX, where `sshd_manage_service` is default *false*, but
 `sshd_allow_reload` is default *true*)
 
+#### sshd_allow_restart
+
+Some changes, for example of the sysconfig and environment files require the full
+restart of the service. If set to *false*, a restart of sshd won't happen on these
+changes. This can help with troubleshooting. You'll need to manually restart sshd
+if you want to apply the changed configuration. Defaults to the same value as
+`sshd_manage_service`.
+
 #### sshd_install_service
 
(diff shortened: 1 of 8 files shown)


Jakuje:
For the record, the CI issues are fixed in #304 so once it will be in, we can rebase this.

handlers/main.yml
```diff
@@ -10,7 +10,19 @@
     - ansible_connection != 'chroot'
     - ansible_facts['os_family'] != 'AIX'
     - ansible_facts['os_family'] != 'OpenWrt'
```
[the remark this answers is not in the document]
richm: never mind

richm (approved):

Jakuje:
(rebased to address the CI issues)

richm (approved):

Jakuje:
Right. This change will have the same issue as the #236, which I already forgot about. Let me have a look into that.

richm (approved):

README.md
```diff
@@ -23,6 +23,12 @@ Ubuntu. This is not the default assigned by this module - it will set
 via simple password. If you need this functionality, be sure to set
 `sshd_PermitRootLogin yes` for those hosts.
 
+**NOTE** The sshd service is reloaded/restarted automatically, only if the role is
+invoked using `roles` keyword. Using `include_role` won't trigger handlers
+as described in
+[official documentation](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_handlers.html#controlling-when-handlers-run).
```
mattwillsher, line 29:
```suggestion
invoked using `roles` keyword. Using `include_role` won't trigger handlers
as described in the Ansible ['taskify includes' proposal](https://github.com/ansible/proposals/issues/136). To work around this, call `meta: flush_handlers` as detailed in the
[official Ansible documentation](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_handlers.html#controlling-when-handlers-run).
```

mattwillsher (commented):
A little clarification around the include_roles matter, up to you if you want to include it. 
Thanks for your work here. Satisfying to see some many issues close in one PR :)

README.md
Jakuje: Accepted. Looks good!

mattwillsher (approved):

mattwillsher:
All good - please merge when ready 👍🏼

Jakuje:
Thank you for comments and reviews! I am not sure if we merge by rebasing/merge commit or squashing to keep the automation working correctly so I will leave the merge on you or Rich.

mattwillsher:
By squash to preserve the `fix:` prefix, I think. @richm does that sound correct?

spetrosi:
The automation works with PR titles, not with commits. Commits can have any format, they should not be consistent with PR titles.
Some time ago we did use commits for automation, but we reverted this to keep commits flexible and developer-oriented, while PRs descriptions are more user-oriented.

Jakuje:
Thanks for clarification. Then I will do rebase & merge to preserve the commits.

PR #308, opened 2025-03-24
fix: service: Add default Environment option

Jakuje:
Enhancement: New Fedora service definition contains definition of empty environment variable used in the Exec command.

Reason: Without that, systemd complains about undefined variable, if the `EnvironmentFile` does not exist or is empty.

Result: The generated Fedora service file is compatible with the system one.

This should not hurt any other distros


richm:
@Jakuje not sure what's going on - but let's first update to the latest stable version of ansible-lint - https://github.com/willshersystems/ansible-sshd/pull/309

templates/sshd.service.j2
```diff
@@ -12,6 +12,9 @@ Documentation=man:sshd(8) man:sshd_config(5)
 
 [Service]
 Type=notify
+{% for var in __sshd_environment_variable %}
+Environment={{ var }}=
```
spetrosi, line 16: Why extra = at the end of the string?

Jakuje:
Thank you for the reviews! Merging to get the Fedora CI back to shape in main!

PR #309, opened 2025-03-24
ci: use ansible-lint v25 - have to use requirements file for ansible-lint now

richm:
This PR supersedes https://github.com/willshersystems/ansible-sshd/pull/306
because there is some extra work that must be done to use v25
We need to provide a requirements file.  Even though sshd does not have
a tests collection-requirements.yml file, we use the same code we
use in linux-system-roles that handles/merges the meta and tests
requirements files.

### DIFF
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -23,16 +23,50 @@ jobs:
       - name: Install tox, tox-lsr
         run: |
           set -euxo pipefail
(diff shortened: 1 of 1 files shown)


Jakuje (approved):
this looks like working better than the current version. Thanks!

Jakuje:
The fedora test will be fixed with #308. I will rebase on top of these changes.

PR #311, opened 2025-06-18
refactor: Ansible 2.19 support

richm:
Ansible 2.19 introduces some big changes
https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_core_2.19.html

A big change is that a boolean expression in a `when` or similar construct
must be converted to a boolean - we cannot rely on the implicit evaluation in
a boolean context.  For example, if `var` is some iterable, like a `dict`, `list`,
or `string`, you used to be able to evaluate an empty value in a boolean context:

```yaml
when: var  # do this only if var is not empty
```

You now have to explicitly test for empty using `length`:

```yaml
when: var | length > 0  # do this only if var is not empty
```

Similarly for `int` values - you cannot rely on `0` being evaluated as false
and non-zero true - you must explicitly compare the values with `==` or `!=`

In macros in templates, the implicit return value is now `none` - we have to
ensure that the macro returns an empty string in these cases - to do this,
use `{{- '' -}}`

The `ansible_managed` variable cannot be overwritten - use a temp variable.

This also fixes some ansible-lint issues with the new ansible-lint

* Task names in the same file must be unique
* Stricter checking for spacing
* Stricter enforcement of the "no quotes in when/that expressions"

These are the biggest changes.  See the porting guide for others.

Signed-off-by: Rich Megginson <rmeggins@redhat.com>

PR #312, opened 2025-06-19
feat: Add new options from OpenSSH 10.0

Jakuje:
Enhancement: Add new options from OpenSSH 10

Reason: The OpenSSH 10.0 provides a new configuration option `SshdAuthPath`.

Result: Users can configure openssh with the new configuration option

### DIFF
--- a/meta/options_body
+++ b/meta/options_body
@@ -106,6 +106,7 @@ SecurityKeyProvider
 SetEnv
 ServerKeyBits
 ShowPatchLevel
+SshdAuthPath
 SshdSessionPath
 StreamLocalBindMask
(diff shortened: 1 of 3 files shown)


Jakuje:
Thank you for the review! I will rather probably wait for @richm to make sure I did not mess up the ostree definitions as I do not know how to run tests on these easily.

richm:
Do you guys want to merge this first or #311?  Either way is fine with me.

Jakuje:
> Do you guys want to merge this first or #311? Either way is fine with me.

I do not have preference. Whatever works for you. But they might cause conflict as they touch the same file so the other will likely need rebase.

Issue #314, opened 2025-07-21
Include Option can not override some options which are configured beforehand


domrim:
At the moment the `Include` option is somewhere in the middle of all options (where it belongs alphabetically).
See https://github.com/willshersystems/ansible-sshd/blob/main/templates/sshd_config.j2#L85 and https://github.com/willshersystems/ansible-sshd/blob/main/templates/sshd_config.j2#L186

This produces the problem that included files can not contain options which were defined before the `Include` option.

I would propose to move the `Include` option to the beginning of each block (inside match and for the body). This would also be in line with defaults from Debian (see also https://manpages.debian.org/bookworm/openssh-server/sshd_config.5.en.html) and AlmaLinux, other OSes should have the same default behaviour.

PR #315, opened 2025-07-22
feat: add Debian 13 support

domrim:
Enhancement: Add support for Debian 13 (aka Trixie)

Reason: 
Like two years ago (#283) the next debian release is near.
Debian 13 Release is planned for 2025-08-09 (https://wiki.debian.org/DebianTrixie)

Result: Support for Debian 13

Issue Tracker Tickets (Jira or BZ if any): -

PR #316, opened 2025-07-22
fix: include external config files first so they can override all options

domrim:
Enhancement: Put the `Include` config option always at the top

Reason: see #314

Result: always load included configs first

### DIFF
--- a/meta/options_body
+++ b/meta/options_body
@@ -1,3 +1,4 @@
+Include
 Port
 AddressFamily
 ListenAddress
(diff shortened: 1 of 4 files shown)


mattwillsher:
Fixing the test sounds like the best approach.

 It will need a minor version bump as it could cause issues if the original config order was masking configuration. #315 will take care of that, so we'll release both together.

@Jakuje @richm can you foresee this change causing any issues?

Jakuje:
> How should we proceed here? adjust tests so they succeed again? remove the file? i would tend to leave the file and adjust the tests accordingly.

Likely yes. If I see right, this test failing (in Fedora at least) is https://github.com/willshersystems/ansible-sshd/blob/main/tests/tests_deprecated_sshd_variable.yml which is not really testing the precedence, but that we can use the deprecated `sshd` variable to set the options. So dropping the ciphers from there or finding another option instead of that should do.

domrim:
this should now be ready for review/merge

mattwillsher (approved):

Issue #307, opened 2025-02-12
Ubuntu 24.04 requires daemon-reload after changing ListenAddress


gilbert-hsu:
On Ubuntu 24.04 (Noble), after add / modify the ListenAddress in sshd config, it is necessary to run `systemctl daemon-reload` to apply the updated systemd unit settings for ssh.socket. Without this step, the changes do not fully take effect because SSH uses socket-based activation.   

Reference:
[SSHD socket-based activation in Ubuntu 22.10 and later ](https://discourse.ubuntu.com/t/sshd-now-uses-socket-based-activation-ubuntu-22-10-and-later/30189)

Would it be possible to update the role to include systemctl daemon-reload when modifying ListenAddress?

gilbert-hsu:
Base on our test

1. How `ssh.service` and `sshd.socket` Work

    `sshd.socket` is managed by systemd, handling the listening addresses and ports.
    `ssh.service` processes SSH connections but does not control listening addresses directly—this is managed by `sshd.socket`.

2. Effects of Modifying `ListenAddress`

    - Adding a new `ListenAddress`
        - Requires `daemon-reload` + `restart sshd.socket` to take effect.
    - Reducing or keeping the same number of  `ListenAddress`
        - Requires `reload ssh`.

## Test Procedure

1. Add ` ListenAddress 10.255.255.129` in `/etc/ssh/sshd_config.d/00-ansible_system_role.conf`:
    ```sh
    $ sudo vi /etc/ssh/sshd_config.d/00-ansible_system_role.conf
    #
    # Ansible managed
    #
    # willshersystems:ansible-sshd

    ListenAddress 10.255.249.175
    ListenAddress 10.255.255.129
    AuthenticationMethods publickey
    ChallengeResponseAuthentication no
    PasswordAuthentication no
    PermitRootLogin no
    PubkeyAuthentication yes
    UsePAM yes
    ```

2. Check the current `ssh.socket` settings:
    ```sh
    $ cat /run/systemd/generator/ssh.socket.d/addresses.conf
    # Automatically generated by sshd-socket-generator

    [Socket]
    ListenStream=
    ListenStream=10.255.249.175:22
    ```

3. Reload `ssh.service` and check `ssh.socket` settings again:
    ```sh
    $ sudo systemctl reload ssh
    $ cat /run/systemd/generator/ssh.socket.d/addresses.conf
    # Automatically generated by sshd-socket-generator

    [Socket]
    ListenStream=
[...]

PR #318, opened 2025-07-29
feat: Support for daemon reload, socket restart and systemd socket file to match Ubuntu 24.04

Jakuje:
Enhancement: Adjust generated systemd socket file to contain actual ports and addresses from the configuration file, add support for the systemctl daemon reload and sshd socket restart.

Reason: The new Ubuntu landed with changed socket file. We also need to reload systemd daemon when we change these, which was reported in #307. 

Result: The generated socket file matches configured port and addresses and systemd daemon is properly reloaded and sshd socket restarted on relevant changes.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,11 +1,23 @@
 ---
 
+- name: Systemd daemon is reloaded when needed
+  ansible.builtin.command: systemctl daemon-reload  # noqa command-instead-of-module
+  # # Can not use systemd_service as it fails on the old Ansible even if not executed
+  # ansible.builtin.systemd_service:
+  #   daemon_reload: true
+  when:
+    - sshd_manage_service | bool
+    - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
+    - ansible_connection != 'chroot'
+  listen: sshd_daemon_reload
+  changed_when: true
+
 - name: Reload the SSH service
   ansible.builtin.service:
     name: "{{ sshd_service }}"
     state: reloaded
   when:
-    - sshd_allow_reload|bool
+    - sshd_allow_reload | bool
     - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
     - ansible_connection != 'chroot'
     - ansible_facts['os_family'] != 'AIX'
@@ -17,7 +29,7 @@
     name: "{{ sshd_service }}"
     state: restarted
   when:
-    - sshd_allow_restart|bool
+    - sshd_allow_restart | bool
     - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
     - ansible_connection != 'chroot'
     - ansible_facts['os_family'] != 'AIX'
@@ -40,7 +52,7 @@
   listen: sshd_reload
   changed_when: false
   when:
-    - sshd_allow_reload|bool
+    - sshd_allow_reload | bool
     - ansible_facts['os_family'] == 'AIX'
 
 # sshd on OpenWrt does not support reloading a service, it has to be restarted instead
@@ -49,6 +61,6 @@
     name: "{{ sshd_service }}"
     state: restarted
   when:
-    - sshd_allow_reload|bool
+    - sshd_allow_reload | bool
     - ansible_facts['os_family'] == 'OpenWrt'
   listen: sshd_reload
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -6,7 +6,6 @@
   # ansible.builtin.systemd_service:
   #   daemon_reload: true
   when:
-    - sshd_manage_service | bool
     - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
     - ansible_connection != 'chroot'
   listen: sshd_daemon_reload
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -36,6 +36,16 @@
     - ansible_facts['os_family'] != 'OpenWrt'
   listen: sshd_restart
 
+- name: Restart the SSH socket
+  ansible.builtin.service:
+    name: "{{ sshd_service }}.socket"
+    state: restarted
+  when:
+    - sshd_allow_restart | bool
+    - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
+    - ansible_connection != 'chroot'
+  listen: sshd_socket_restart
+
 # sshd on AIX cannot be 'reloaded', it must be Stopped+Started.
 # It's dangerous to do this in two tasks.. you're stopping SSH and then trying to SSH back in to start it.
 # Instead, use a dirty shell script:
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -18,6 +18,7 @@
     state: reloaded
   when:
     - sshd_allow_reload | bool
+    - sshd_systemd_unit == 'service'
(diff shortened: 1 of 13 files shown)


templates/sshd.socket.j2
```diff
@@ -8,7 +8,18 @@ Before=sockets.target
 {% endif %}
 
 [Socket]
-ListenStream=22
+{% for port in __sshd_ports_from_config %}
+{%  if __sshd_listen_addresses_from_config != '["0.0.0.0","[::]"]' or __sshd_socket_listen_expand %}
+{%   for addr in __sshd_listen_addresses_from_config %}
```
richm, line 13: I think `__sshd_listen_addresses_from_config` will always be a `string` and not a `list`?  In the above code, every code path results in a `to_json` which is a `string`

Jakuje: Thanks for the prompt look into this! Yes, this will be the issue. Will  need to fix!

Jakuje: Sounds like it was the one. Fixed now. Lets see what CI will say

handlers/main.yml
```diff
@@ -1,5 +1,17 @@
 ---
 
+- name: Systemd daemon is reloaded when needed
+  ansible.builtin.command: systemctl daemon-reload  # noqa command-instead-of-module
+  # # Can not use systemd_service as it fails on the old Ansible even if not executed
+  # ansible.builtin.systemd_service:
+  #   daemon_reload: true
+  when:
+    - sshd_manage_service|bool
```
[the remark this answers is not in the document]
Jakuje: Fixed in other handlers too. Thanks !

Jakuje:
Ugh. Green. Works in containers. @richm how do you invoke the tests on real VMs to make sure the reload really does something?

Jakuje:
Added one more change to actually address 307 (hopefully). The local run in centos 10 VM worked for me.

spetrosi (approved):
Does this require updating README.md or this is a fix that is not visible to users?

tasks/install_config.yml
```diff
@@ -25,6 +25,16 @@
...
   notify: sshd_reload
+  register: __sshd_config_result
+
+- name: Notify systemd to reload daemon when configuration changes and we have non-standard address/port
+  ansible.builtin.assert: { that: true, quiet: true }
+  changed_when: true
+  when:
+    - __sshd_config_result is changed
+    - not __sshd_socket_accept
+    - __sshd_listen_addresses_from_config | trim != '["0.0.0.0", "[::]"]' or __sshd_ports_from_config | trim != "[22]"
```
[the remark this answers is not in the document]
Jakuje: Yes, but we do not have a simple way to get the previous port/address configuration as far as I know so I did here "reload if any non-default port or address is used and configuration as a whole is changed", which might cause some unnecessary reloads. Not sure if there is better way to do this.

Jakuje:
> Does this require updating README.md or this is a fix that is not visible to users?

Good point. I think we should update also readme as this is a change users might notice as it reloads the systemd daemon. And we should likely add `sshd_allow_daemon_reload` for users not wanting to reload a systemd daemon for whatever reason.

tasks/install_config.yml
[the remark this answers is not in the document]
Jakuje: Thats basically what we get in `__sshd_listen_addresses_from_config` and `__sshd_listen_ports_from_config` -- if it was not specified in the playbook, it will use the defaults which are tested here.

But that is not true -- if the previous configuration have these specified and new one does not have, the new configuration will be changed to default, but this reload will not be triggered as there are no ports/addresses in the playbook.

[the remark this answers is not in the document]
richm: ok - I guess this is the best we can do unless we implement some sort of sshd config parser

Jakuje:
> @Jakuje ready to merge?

I got some more comments in #307, that I did not get back to as it looks like it creates something wrongly. So please wait. I can split the fixes to separate PR if you want.

Jakuje:
@richm is this something the old ansible on RHEL6 and RHEL7 does not like in the new template? What I did was just adding the whitespace control ...

richm:
> @richm is this something the old ansible on RHEL6 and RHEL7 does not like in the new template? What I did was just adding the whitespace control ...

Something weird is going on - investigating

Jakuje:
> > @richm is this something the old ansible on RHEL6 and RHEL7 does not like in the new template? What I did was just adding the whitespace control ...
> 
> Something weird is going on - investigating

I think the only change I did was this:

https://github.com/willshersystems/ansible-sshd/compare/26bcea52d18e3f8be635ce0a5208abe7f7203766..9ffe33618757a91727205bfb58fa70017037d29d

Let me revert the indentation inside braces. I rewrite the indent into the `{%` in the attempt to avoid the whitespace, but then I also used the minus sign (as it really did not help).

tasks/find_bind_addresses.yml
```diff
@@ -0,0 +1,22 @@
...
+      {%- elif __sshd_config['ListenAddress'] is defined -%}
+        {{- __sshd_config['ListenAddress'] | to_json -}}
+      {%- elif __sshd_defaults['ListenAddress'] is defined and not sshd_skip_defaults -%}
+        {{- __sshd_defaults['ListenAddress'] | to_json -}}
+      {%- else -%}
+        {{- ["0.0.0.0", "[::]"] | to_json -}}
+      {%- endif -%}
+  ansible.builtin.set_fact:
+    __sshd_listen_addresses_from_config: >-
+      {%- if __sshd_listen_from_config_tmp | from_json is string or __sshd_listen_from_config_tmp | from_json is number -%}
```
richm, line 18:
```suggestion
      {%- if __sshd_listen_from_config_tmp  | type_debug == "list" -%}
      {{-     __sshd_listen_from_config_tmp | to_json %}
      {%- elif __sshd_listen_from_config_tmp | from_json is string or __sshd_listen_from_config_tmp | from_json is number -%}
```

tasks/find_ports.yml
```diff
@@ -4,23 +4,19 @@
...
-      {% if __sshd_ports_from_config_tmp | from_json is string or __sshd_ports_from_config_tmp | from_json is number %}
-        {{ [__sshd_ports_from_config_tmp | from_json] | to_json }}
-      {% else %}
-        {{ __sshd_ports_from_config_tmp }}
-      {% endif %}
-  when:
-    - sshd_manage_firewall | bool or sshd_manage_selinux | bool
-    - ansible_facts['os_family'] == 'RedHat'
-    - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
+      {%- if __sshd_ports_from_config_tmp | from_json is string or __sshd_ports_from_config_tmp | from_json is number -%}
```
richm, line 18:
```suggestion
      {%- if __sshd_ports_from_config_tmp | type_debug == "list" -%}
      {{-    __sshd_ports_from_config_tmp | to_json -%}
      {%- elif __sshd_ports_from_config_tmp | from_json is string or __sshd_ports_from_config_tmp | from_json is number -%}
```

richm: There's some sort of odd type conversion going on in ansible 2.9 - for some reason the `to_json` is being ignored and the original `list` type is being preserved.  This adds a condition to check for this.

Jakuje: Thanks for having a look into that! Pushed the change. Lets see.

mattwillsher:
Looks good to me. Thanks for your work on this @Jakuje

PR #320, opened 2025-08-01
docs(changelog): version v0.27.1 [citest skip]

richm:
Update changelog and .README.html for version v0.27.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+[v0.27.1] - 2025-08-01
+--------------------
+
+### Bug Fixes
+
+- fix: New configuration option in CentOS 10 (#319)
+
 [v0.27.0] - 2025-06-19
 --------------------
 


mattwillsher:
Change includes more than just Centos 10 changes? What's the cause of the Ubuntu latest failure? May it cause issues for end users?

Jakuje:
> Change includes more than just Centos 10 changes? 

Yes, I squeezed also removal of the Debian buster CI as its EOL and repositories no longer available. And a debug output in the test. Neither is user-visible.

> What's the cause of the Ubuntu latest failure? May it cause issues for end users?

The ubuntu failure is just a test generating systemd socket file (I assume most of people do not use this). And as Ubuntu changes these things mid-release, its hard to keep up with them. It is fixed in #318, but not completely solving #307 so I will need to revisit this some time later when I will get back to that, hopefully during next days, rather than merging something half-baked.

Jakuje (approved):

PR #330, opened 2025-10-08
fix: Allow specifying OS vars from playbook_dir

lkirkwood:
Enhancement: Previously, the "first_found" lookup would always find the `role_path/vars/main.yml` file before any `playbook_dir/vars/...` file, so it was impossible to set variables from the second directory.

This change:
+ Removes `main.yml` from the lookup file names so that the automatically loaded `role_path/vars/main.yml` file can never stop another file from being loaded
+ Replaces the inline `lookup('first_found')` call with a `with_found_first` block so that the task won't fail if there is no additional vars file.

Result: The behaviour of the role should be identical, except that it is now possible to use the `{{ playbook_dir }}/vars/...` files that the role clearly intends to support.

### DIFF
--- a/tasks/variables.yml
+++ b/tasks/variables.yml
@@ -18,7 +18,7 @@
         __sshd_is_ostree: "{{ __ostree_booted_stat.stat.exists }}"
 
 - name: Set OS dependent variables
-  ansible.builtin.include_vars: "{{ lookup('first_found', params) }}"
+  ansible.builtin.include_vars: "{{ item }}"
   vars:
     _distribution_lts_offset: >-
       {{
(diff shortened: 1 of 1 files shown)


mattwillsher:
Thanks @lkirkwood

infinite-ops:
I think this actually fixes an issue with OpenBSD in 0.28.0 which has been broken since 065812b3, OpenBSD does not have a 'distribution_major_version' .

I was going to PR it, until I found that this PR, the issue is gone. This was my change:
```
diff --git a/roles/willshersystems.sshd/tasks/variables.yml b/roles/willshersystems.sshd/tasks/variables.yml
index aef04786..42dba69d 100644
--- a/roles/willshersystems.sshd/tasks/variables.yml
+++ b/roles/willshersystems.sshd/tasks/variables.yml
@@ -20,15 +20,21 @@
 - name: Set OS dependent variables
   ansible.builtin.include_vars: "{{ lookup('first_found', params) }}"
   vars:
+    _distribution_major_version: >-
+      {{
+        ansible_facts['distribution_version'].split('.')[0]
+        if ansible_facts['distribution'] == "OpenBSD"
+        else ansible_facts['distribution_major_version']
+      }}
     _distribution_lts_offset: >-
       {{
-        ansible_facts['distribution_major_version'] | int % 2
+        _distribution_major_version | int % 2
         if ansible_facts['distribution'] == "Ubuntu"
         else 0
       }}
     _distribution_lts_version: >-
       {{
-        ansible_facts['distribution_major_version'] | int -
+        _distribution_major_version | int -
         _distribution_lts_offset | int
         if ansible_facts['distribution'] == "Ubuntu"
         else ansible_facts['distribution_version']
@@ -36,7 +42,7 @@
     params:
       files:
         - "{{ ansible_facts['distribution'] }}_{{ _distribution_lts_version }}.yml"
-        - "{{ ansible_facts['os_family'] }}_{{ ansible_facts['distribution_major_version'] }}.yml"
+        - "{{ ansible_facts['os_family'] }}_{{ _distribution_major_version }}.yml"
```

PR #335, opened 2025-11-14
ci: pin gh actions inline with security practices

mattwillsher:
Enhancement: Pin GHA actions to sha in line with good practice

Reason: Avoid unexpected changes to actions.

Result:


richm:
Will be interesting to see if dependabot can keep up with the changes

PR #356, opened 2026-03-20
test: ensure role gathers the facts it uses by having test clear_facts before include_role

richm:
The role gathers the facts it uses.  For example, if the user uses
`ANSIBLE_GATHERING=explicit`, the role uses the `setup` module with the
facts and subsets it requires.

This change allows us to test this.  Before every role invocation, the test
will use `meta: clear_facts` so that the role starts with no facts.

Create a task file tests/tasks/run_role_with_clear_facts.yml to do the tasks
to clear the facts and run the role.  Note that this means we don't need to
use `gather_facts` for the tests.

Some vars defined using `ansible_facts` have been changed to be defined with
`set_fact` instead.  This is because of the fact that `vars` are lazily
evaluated - the var might be referenced when the facts have been cleared, and
will issue an error like `ansible_facts["distribution"] is undefined`.  This is
typically done for blocks that have a `when` condition that uses `ansible_facts`
and the block has a role invocation using run_role_with_clear_facts.yml
These have been rewritten to define the `when` condition using `set_fact`.  This
is because the `when` condition is evaluated every time a task is invoked in the
block, and if the facts are cleared, this will raise an undefined variable error.

### DIFF
--- a/.github/workflows/ansible-centos-check.yml
+++ b/.github/workflows/ansible-centos-check.yml
@@ -26,6 +26,9 @@ jobs:
         tests/tests_duplicate_role.yml
         tests/tests_os_defaults.yml
         tests/tests_firewall_selinux.yml
+    - run: >-
+        sed -i -e '/public: /d'
+        tests/tasks/run_role_with_clear_facts.yml
     - run: "sed -i -e 's/ansible.builtin.//g' -e 's/ansible.posix.//g' */*.yml */*/*.yml"
 
     - name: ansible check with centos 6
--- a/tasks/variables.yml
+++ b/tasks/variables.yml
@@ -1,9 +1,11 @@
 ---
 - name: Ensure ansible_facts used by role
   ansible.builtin.setup:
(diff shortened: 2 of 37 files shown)


richm:
@Jakuje looks like there are some new config settings in centos-10

NOTE: The reason I changed the fact gathering condition to use `difference` instead of `intersect` is that with the latter, the comparison of the present facts with the required facts is being done on unsorted lists.  The comparison may fail if the only difference is the order.  This means that facts are gathered unnecessarily.  Using `difference`  works no matter what the order is.

Jakuje:
> @Jakuje looks like there are some new config settings in centos-10

Sounds like we have. Will add separare PR to add them. I thought the solution was something outside of OpenSSH.

Jakuje (approved):
otherwise lgtm

Jakuje (approved):
lets resolve the remaining centos issue once we will have the build fixing the manual page separately.

PR #359, opened 2026-04-22
chore: adding the default include for leap

kees-closed:
Enhancement: The default include line is removed when this role is executed.

Reason: I suppose it is handy to keep it, since the distro uses it to set crypto policies.

Result:
In a `podman run -it docker.io/geerlingguy/docker-opensuseleap15-ansible:latest bash` container, I ran these commands after installing sshd.

```
9e8a3eb0c81e:/ # grep -r Include /etc/ssh/*
/etc/ssh/sshd_config:Include /etc/ssh/sshd_config.d/*.conf
/etc/ssh/sshd_config:Include /usr/etc/ssh/sshd_config.d/*.conf
/etc/ssh/sshd_config.d/40-suse-crypto-policies.conf:Include /etc/crypto-policies/back-ends/opensshserver.config

9e8a3eb0c81e:/ # cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="15.6"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.6"
PRETTY_NAME="openSUSE Leap 15.6"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.6"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
DOCUMENTATION_URL="https://en.opensuse.org/Portal:Leap"
LOGO="distributor-logo-Leap"

9e8a3eb0c81e:/ # rpm -qc openssh-server
/etc/pam.d/sshd
/etc/slp.reg.d/ssh.reg
/etc/ssh/sshd_config
/etc/ssh/sshd_config.d/40-suse-crypto-policies.conf
/etc/sysconfig/SuSEfirewall2.d/services/sshd
```

This include line is there, but not in the config itself for LEAP.

### DIFF
--- a/vars/openSUSE Leap_15.yml
+++ b/vars/openSUSE Leap_15.yml
@@ -3,6 +3,7 @@ __sshd_packages:
   - openssh
 __sshd_sftp_server: /usr/lib/ssh/sftp-server
 __sshd_defaults:
+  Include: /usr/etc/ssh/sshd_config.d/*.conf
   AuthorizedKeysFile: .ssh/authorized_keys
   UsePAM: true
   X11Forwarding: true


Jakuje:
the opensuse configuration is from 2019 without much change as we do not have a CI to run tests properly on that: 6be10a2d17c3ed3a1d57557a10b47ca6d8260a7d

If you care for the opensuse, it would be helpful if some simple CI pipeline could be set up. We are using GH actions from

https://github.com/roles-ansible/check-ansible-fedora-latest-action

but they are quite simple and some ad-hoc solution for your container could be enough.

Jakuje (approved):

Jakuje:
Thank you! I will merge your changes and finalize the CI for leap and tumbleweed in #367

PR #360, opened 2026-04-23
feat: add Ubuntu 26 LTS (Resolute Raccoon) support

sbourdette:
Add vars/Ubuntu_26.yml with the same configuration as Ubuntu 24 LTS: socket-based activation, drop-in directory structure under /etc/ssh/sshd_config.d/. Also add resolute to supported platforms in meta/main.yml.

Enhancement:

Reason:

Result:

PR #361, opened 2026-04-27
feat: add role fingerprints to syslog

richm:
Feature: Add a fingerprint string to the system log to indicate when the role began
successfully, and when the role finished successfully.  The fingerprint string indicates
the role name, a timestamp, and the platform.

Reason: Users can see when the role was used and if it was used successfully.  This
information from the system log can be collected by log scanners and aggregators
for further analysis.

Result: The role logs fingerprints to the system log.

This also adds a test to check if the fingerprints were written upon a successful
role invocation.


richm:
Q: What is the motivation for this?
A: system logs are gathered by customers and sent in to Red Hat support for triage/debugging.  The fingerprint helps us know 1) what roles were used 2) when they were used 3) if the role was successful 4) some additional metadata.

Q: Why not just use the "logger" command or whatever is the command that logs to the system logs?
A: logger command is provided by the util-linux or util-linux-core package - new package dependencies for every role, platform differences

Q: Why not use the "syslog" python library directly?  Why use the Ansible `module.log` function?
A: The `module.log` function handles some corner cases we would need to add to our module.  We don't need to reference the "syslog" library at all.

Jakuje (approved):
lgtm! The changes to lint are fixes for the issues I can see in #360?

richm:
> lgtm! The changes to lint

not sure what you mean - the added `.sanity-*` files are for ansible-test - it really wants python modules to use `GPLv3`, so you have to tell `ansible-test` to not complain about modules that use other licenses.

> are fixes for the issues I can see in #360?

That looks unrelated to this PR.

Jakuje:
Oh, they are probably related to update to Fedora 44. I started the pipeline just today. So this PR is good and we need to figure out why the Fedora 44 does not work.

Jakuje:
Fix for Fedora GH action is here:

https://github.com/roles-ansible/check-ansible-fedora-latest-action/pull/7

richm:
> > lgtm! The changes to lint
> 
> not sure what you mean - the added `.sanity-*` files are for ansible-test - it really wants python modules to use `GPLv3`, so you have to tell `ansible-test` to not complain about modules that use other licenses.
> 
> > are fixes for the issues I can see in #360?
> 
> That looks unrelated to this PR.

Also note that the `.sanity-*` files are only used (for now, anyway) for the fedora.linux_system_roles collection - part of the process of building/bundling the roles into the collection will merge all of the various `.sanity-*VER.txt` files into the collection `tests/sanity/ignore-VER.txt` files.

If this role is ever converted into a standalone collection, you can simply copy the files into `tests/sanity/ignore-VER.txt`

PR #366, opened 2026-05-11
feat: Add verbosity-based logging to reduce verbose output

spetrosi:
Enhancement: Add verbosity-based logging using \`no_log: "{{ ansible_verbosity < 3 }}"\` to reduce verbose output from fact gathering and service management tasks.

Reason: The setup task (fact gathering) and service task (enabling/starting sshd) can produce verbose output that clutters logs during normal playbook execution. This output is useful for debugging but creates noise in standard runs.

Result: By using \`ansible_verbosity < 3\`, the output from these tasks is suppressed by default but can be shown when running with \`-vvv\` or higher verbosity for debugging purposes. This improves log readability while maintaining full debugging capabilities when needed.

Note: This is for verbosity control, not for security purposes - there's no sensitive information being hidden, just reducing noise in normal operation.

### DIFF
--- a/tasks/install_service.yml
+++ b/tasks/install_service.yml
@@ -51,6 +51,7 @@
     - sshd_manage_service|bool
     - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
     - ansible_connection != 'chroot'
+  no_log: "{{ ansible_verbosity < 3 }}"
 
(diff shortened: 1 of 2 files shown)


richm (approved):

spetrosi:
@Jakuje please review

Jakuje (approved):

PR #367, opened 2026-05-14
ci: Use generalized action to test against different distributions

Jakuje:
Enhancement: Use single GH action to test on container images

Reason: The fleet of GH actions testing against different container images was not maintainable.

Result: On this role side, all code is in one workflow file.


Jakuje:
The new GH Action used here:

https://github.com/marketplace/actions/check-ansible-action

Would be grateful for sharing, using, providing feedback, inputs and stars!

mattwillsher (approved):

PR #369, opened 2026-06-11
test: ensure fips_enabled bind mount is removed from fstab

richm:
The test was leaving the fips_enabled bind mount in /etc/fstab.
Use `state: absent` to both unmount the path and remove from fstab.
https://docs.ansible.com/projects/ansible/latest/collections/ansible/posix/mount_module.html#parameter-state


spetrosi (approved):

spetrosi (approved):

Jakuje (approved):
good catch!

PR #370, opened 2026-06-23
docs(changelog): version v0.33.0 [citest_skip]

richm:
Update changelog and .README.html for version v0.33.0


Jakuje (approved):
the suse CI is broken, but it can be resolved separately.

PR #373, opened 2026-07-21
test: have to use yum for EL7 tests

richm:
Using `dnf` here causes test failures on EL7.  Revert to use `pkg_mgr`
in order to use `yum` on EL7, `dnf` elsewehre.

Update tox-lsr to 3.20.0 to fix issues with tox 4.58

Use `ansible_facts["pkg_mgr"]` instead of `pkg_mgr`

Do not install `fedora.linux_system_roles` on platforms that do not use it for testing

### DIFF
--- a/.github/workflows/ansible-check.yml
+++ b/.github/workflows/ansible-check.yml
@@ -10,20 +10,33 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os_image:
-          - alpine:latest
-          - quay.io/centos/centos:stream8
-          - quay.io/centos/centos:stream9
-          - quay.io/centos/centos:stream10
-          - debian:bullseye
-          - debian:bookworm
-          - debian:latest
-          - fedora:latest
-          - ubuntu:22.04
-          - ubuntu:24.04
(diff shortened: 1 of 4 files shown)


spetrosi (approved):

.github/workflows/ansible-check.yml
```diff
@@ -10,20 +10,33 @@ jobs:
...
+          - os_image: ubuntu:22.04
+            req_file: tests/requirements-no-fedora-lsr.yml
+          - os_image: ubuntu:24.04
+            req_file: tests/requirements.yml
+          - os_image: ubuntu:latest
+            req_file: tests/requirements.yml
+          # - os_image: opensuse/tumbleweed:latest # too odd
+          #   req_file: tests/requirements.yml
+          - os_image: opensuse/leap:latest
+            req_file: tests/requirements-no-fedora-lsr.yml
```
Jakuje: the leap is now broken with ansible too so I would not bother with that and just remove it or comment it out to make a CI happy.

Jakuje (approved):

PR #376, opened 2026-08-03
Add better support for Debian testing/unstable

Zugschlus:
In Debian, distribution_major_version is "forky/sid" for the development versions of the distribution. that makes the code search for a variable file called sid.yml in a subdirectory called "forky" of the directory in the search part.

This PR fixes this by replacing slashes in distribution_major_version with dashes, and adds vars/Debian_forky-sid.yml as demonstration.


Jakuje:
Could you also add the sid here so we can run CI against that?

https://github.com/willshersystems/ansible-sshd/blob/main/.github/workflows/ansible-check.yml#L22

I am not sure if I tried that or why I did not add it in the first place.

Zugschlus:
Debian sid is a quickly moving target and users are not suposed to use that in production. Are you sure you want to run your CI against that?

Zugschlus:
I also must apologize (as a Debian Developer) that we require this kind of special casing for our development versions. There is a heated ongoing discussion about that, involving both the maintainer of our base-files package (which contains the IMO inconsistent data) and our release team (which is very eager to not call testing and unstable a release and to not give it a number before we are actually preparing the release).

In the mean time I have learned that it is possible to override distribution_major_version from the inventory and am doing this for my sid hosts now. This is needed for the change suggested by this PR :-(.

fyi, this is what Debian sid ships in /etc/os-release:
```
PRETTY_NAME="Debian GNU/Linux forky/sid"
NAME="Debian GNU/Linux"
VERSION_CODENAME=forky
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
and this is what we ship in current stable:
```
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.6
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

as far as I can see, this is imported pretty verbatim into ansible's distribution_* facts.

Jakuje:
I frequently say the untested code is broken code. If we do not run this in CI, we will not notice the naming will change and we will not be able to fix it until it breaks somewhere (likely not in production though, but I know things tend to break in the most inconvenient times). I also find useful to be able to "fork" the sid once the final version of debian stabilizes and people come here asking for its support, instead of trying to figure out all what changed since the last release.

Zugschlus:
As a Debian developer, I'd recommend running CI against Debian testing as this is the distribution that you can be more confidential that there are no major breakages like uninstallable packagets etc. When you run your CI against Debian sid, you will have lots of failures that are caused by breakage on the Debian side that we will stuggle to fix.

One can debate running CI against Debian testing, but I'd recommend enabling that only once Debian has frozen to prepare a release. The distribution metadata usually changes a few weeks before the actual release, so maybe it is ia good idea to, for example, starting to test against the next  Debian testing "forky" once the version number 14 has appeared in distribution metadata. Maybe you can even automate this. I don't think it is worth doing automatic testing against a quickly moving target before the freeze, you're ending up with lots of fixes you'll never actually need for a released Debian.

I might be one of the few people running your role against actual systems running Debian unstable. I have, for the time being, overridden the distribution metadata with a group variable so that this change is not needed, so feel free to close this if you feel more comfortable that way. I am available to help with adapting the role to Debian though.

Jakuje (approved):
ok then. Lets merge it as it is.

PR #378, opened 2026-08-03
Change systemd unit status according to sshd_systemd_unit

Zugschlus:
The current role will enable ssh.service unconditionally even if the user has requested to use systemd socket activation. I think this is a bug.

This commit will fix that, keeping proper order: first deactivate and mask the unwanted activation type, then unmask and activate the desired activation type.

The actual transition is not done since this might infuence the ssh session that ansible is running over, probably leaving the machine without a running sshd. It is advised to reboot after changing the configuration.

For the chroot case, only "service" is supported. I don't have the possibility to test this (and also not the time).

There are no automatic tests for this. I don't know how to write them. If somebody helps me with that, I might put some work into this.

It might be a good idea to have an additional task that might print a warning (or error out?) after a prepared transition, reminding the local admin to do the reboot. Please indicate how you'd want to handle that.


tasks/install_service.yml
```diff
@@ -42,13 +42,28 @@
         - sshd_reload
         - sshd_socket_restart
 
-- name: Service enabled and running
-  ansible.builtin.service:
-    name: "{{ sshd_service }}"
+- name: Disable (not stop) unwanted SSH systemd unit
```
Jakuje: I would rewrite the names to declarative tone:
```suggestion
- name: Ensure unwanted SSH systemd unit is disabled
```
Ideally with some comment describing why we do this (enabling both socket and service breaks the setup altogether).

tasks/install_service.yml
```diff
@@ -42,13 +42,28 @@
...
+    _sshd_unit_suffix: >-
+      {{ "socket" if sshd_systemd_unit == 'service'
+          else "service" }}
+  when:
+    - sshd_manage_service | bool
+    - ansible_facts['virtualization_type'] | default(None) not in __sshd_skip_virt_env
+    - ansible_connection != 'chroot'
+  no_log: "{{ ansible_verbosity < 3 }}"
+
+- name: Enable (not start) selected SSH systemd unit
```
Jakuje:
```suggestion
- name: Ensure selected SSH systemd unit is enabled
```

tasks/install_service.yml
```diff
@@ -42,13 +42,28 @@
...
+  ansible.builtin.systemd:
+    name: "{{ sshd_service }}.{{ sshd_systemd_unit }}"
     enabled: true
```
Jakuje, line 66: Good catch! Given that this code is not executed in containers, its quite tough to test it here in CI. Do you have some idea how this could be tested without spinning up the whole VMs?

Zugschlus: Sadly, I dont know enough about tox and your test framework to comment here. I have not been able (yet) to run the testsuite of this role successfully (or even get it to begin!). Maybe some kind of closer discussion (like IRC, a Mailing List or similar) might help here.

As far as i know, it should be possible to start containers with their own service manager, that way a full test could be done, otherwise systemctl could be mocked and wrapped so that we can see whether the role issues the intended calls.

tasks/install_service.yml
Zugschlus: done. thanks for the guidance.

tasks/install_service.yml
Zugschlus: done as well.

tasks/install_service.yml
Jakuje: Yeah, we could run the containers with systemd, but thats awfully more complicated than running containers as they are meant to be so that was one of the things I wanted to avoid. I guess we do not need to have test specifically for this.

Jakuje (approved):

Zugschlus:
Is there anything to be done on my side here?

Jakuje:
> Is there anything to be done on my side here?

I do not think so. I just approved and forgot about the PR. Sorry. 

The bullseye CI sounds broken now (its actually eol, which might be the reason so I will drop it from the CI with next PR).

I will merge it once the CI will pass (except for the known issues)

PR #379, opened 2026-08-04
add ansible managed comments to systemd unit templates

Zugschlus:
Enhancement:
add ansible managed comments to the systemd unit templates so that they can be
recognized on the target system as coming from this role


templates/auth_principals.j2
```diff
@@ -1,5 +1,5 @@
-{{ ansible_managed | comment }}
-{{ "willshersystems:ansible-sshd" | comment(prefix="", postfix="") }}
+{{ ansible_managed | comment(prefix="", postfix="") -}}
```
richm, line 1: why `(prefix="", postfix="") -` here?

templates/auth_principals.j2
```diff
@@ -1,5 +1,5 @@
...
+{{- "willshersystems:ansible-sshd" | comment(prefix="", postfix="") }}
```
richm, line 2: why `{{-` here?

Zugschlus: because it will add empty lines otherwise.

Zugschlus:
I am closing this. I didnt realize some of the templates are generated. Will come back to this later.

Zugschlus:
That one is not going to come back, there is a test that test for this very formatting of the ansible-managed marking, so somebody wanted it to be exactly this way.

PR #380, opened 2026-08-05
Modify sshd_config templates to properly handle lists

Zugschlus:
This is the second try to allow configuration lists like
```
        sshd_config:
          HostKeyAlgorithms:
            - "ssh-ed25519-cert-v01@openssh.com"
            - "ssh-ed25519"
          KexAlgorithms:
            - "curve25519-sha256"
          Ciphers:
            - "chacha20-poly1305@openssh.com"
            - "aes256-gcm@openssh.com"
          MACs:
            - "hmac-sha2-512-etm@openssh.com"
          HostbasedAcceptedAlgorithms:
            - "ecdsa-sha2-nistp256"
            - "ecdsa-sha2-nistp384"
            - "ecdsa-sha2-nistp521"
            - "ssh-ed25519"
```
to render into sshd config correctly while preserving backwards compatibility.

The list of options where this is enabled (HostKeyAlgorithms, KexAlgorithms, Ciphers, MACs, HostbasedAcceptedAlgorithms, PubkeyAcceptedAlgorithms, CASignatureAlgorithms, and a few others) is most probably not exhaustive. Please indicate when you want me to sift through the options lists to find options where this behavior may be appropriate. It surely isnt for example, for AcceptEnv, AllowUsers, or Match.

This patch modifies all: the metatemplate, the option list files (introducing two new files for options that support a list), the template generator and it also includes freshly included templates.


richm:
@Zugschlus where is the list format documented?  A quick web search doesn't reveal anything except the sshd_config man page which does not go into these details.  This role has been heavily used for several years - I am really surprised no one has had a problem with this until now.

> The list of options where this is enabled (HostKeyAlgorithms, KexAlgorithms, Ciphers, MACs, HostbasedAcceptedAlgorithms, PubkeyAcceptedAlgorithms, CASignatureAlgorithms, and a few others) is most probably not exhaustive.

How are you determining this?  Can this process be automated?  For example, the test https://github.com/willshersystems/ansible-sshd/blob/main/tests/tests_all_options.yml#L92 parses some of the option information out of the man page.

Zugschlus:
Hi.

The current version of the role does already allow settings like
HostKeyAlgorithms: "ssh-ed25519-cert-v01@openssh.com,ecdsa-sha2-nistp256-cert-v01@openssh.com,ecdsa-sha2-nistp384-cert-v01@openssh.com,ecdsa-sha2-nistp521-cert-v01@openssh.com". So setting this configuration options is already possible for current users of the role, which might answer your first question.

Allowing a list here makes the input data more ansiblish, and takes away the burden of having to find out whether the software wants the list space or comma separated. Some configuration options want the one, others want the other (so this needs additional work).

Additionally, there are list options that have include and exclude prefixes, for example the "Ciphers" option. That also needs additional work to be really nice.

I doubt that this can be automated, you'd have to parse the manual page for that, which already uses different wordings for the same thing at different places ("space separated" vs "separated by spaces").

But the sshd_config template needs manual work anyway with each OpenSSH release as the option list is already "hardcoded" there. I do see that this PR probably only goes half the way, but I think it goes in the right direction, and it is backwards compatible.

richm:
> Hi.
> 
> The current version of the role does already allow settings like HostKeyAlgorithms: "[ssh-ed25519-cert-v01@openssh.com](mailto:ssh-ed25519-cert-v01@openssh.com),[ecdsa-sha2-nistp256-cert-v01@openssh.com](mailto:ecdsa-sha2-nistp256-cert-v01@openssh.com),[ecdsa-sha2-nistp384-cert-v01@openssh.com](mailto:ecdsa-sha2-nistp384-cert-v01@openssh.com),[ecdsa-sha2-nistp521-cert-v01@openssh.com](mailto:ecdsa-sha2-nistp521-cert-v01@openssh.com)". So setting this configuration options is already possible for current users of the role, which might answer your first question.

I see.  So if the user knows that the setting takes a comma delimited list, they can pass in a value like `ssh_HostKeyAlgorithms: ecdsa-sha2-nistp256,ecdsa-sha2-nistp384` 

> 
> Allowing a list here makes the input data more ansiblish, and takes away the burden of having to find out whether the software wants the list space or comma separated.

+1

> Some configuration options want the one, others want the other (so this needs additional work).

This is a strength of Ansible roles - being able to abstract and hide these implementation details so that users don't have to worry about them.

> 
> Additionally, there are list options that have include and exclude prefixes, for example the "Ciphers" option. That also needs additional work to be really nice.
> 
> I doubt that this can be automated, you'd have to parse the manual page for that, which already uses different wordings for the same thing at different places ("space separated" vs "separated by spaces").
> 
> But the sshd_config template needs manual work anyway with each OpenSSH release as the option list is already "hardcoded" there. I do see that this PR probably only goes half the way, but I think it goes in the right direction, and it is backwards compatible.

Maybe @Jakuje can suggest a way to determine which settings take a list, and if that list is comma or space delimited.

Jakuje:
Generally looks good, but please, provide some tests in puts in the tests/ directory to demonstrate both of the formats work as expected (see existing tests for inspiration). If this should be something supported, this syntax should also be described at least in README. We also describe building the option_lists from two files there. With 2 more additional files, I think the README should be adjusted to bring some more light into this.

Zugschlus:
Hi. After thinking about this for way too much time, I think that the current form of meta/make_option_lists as a shell script doesn't scale. sshd_config has options that take a space separated list, it has options that take a comma separated list, it has options that you write multiple times for multiple values, and it has the nightmare "AuthenticationMethods" that takes a space separated list of comma separated lists. Doing this in the shell script would mean having multiple temporary files and also lots of duplicated code.

If you indicate that such a patch request would have a chance of being accepted, I would like to suggest replacing make_option_lists a small python program that reads a yaml configuration file which in turn classifies the sshd_config options in the classes mentioned above.

Sadly, I do not see any possibility to create this in an automated way, The sshd_config manual page is good to read for a human, but quite hard to parse by a program, and I dont think that upstream ssh would particularly enjoy ansible roles to parse their man pages. This is probably going to remain manual work (it probably is already, isn't it?).

Let me know whether this work has a change not going down the drain and I'll come back to the topic.

Regarding tests, I need some guidance here. I have not been able to get tox working at all, I even installed a Fedora test system to see whether this works, but no success.  Are there any more instructive docs than contributing.md and the linked contribute.html on linux-system-roles.github.io? My "phone jokers" in python development just say "tox is outdated, get them to move to uv and molecule", but I understand that this is most probably not going to happen. Is there any possibility to talk to one of you guys outside of github issues and prs?

Jakuje:
I agree that sshd_config syntax is pain and managing all of its complexity in shell scripts is not something to be readable (personally maintaining one re-implementation I find enough). Python script might be an option, but its a question whether its really worth it to be able to achieve something that is already possible to achieve with different syntax.

Indeed adding a new options is a manual process. The only aid is the CI test that fails when new option is found (with some quirks) from manual pages. But it was mostly reliable over the years.

How to run tests with linux system roles is described in here:

https://linux-system-roles.github.io/contribute.html

But frankly, I have not been much fan of this approach and most of the time run the tests in container, mostly mimicking what the following GH Action does: https://github.com/Jakuje/check-ansible-action/ so generally starting a container, mounting the source in the container, installing ansible, creating simple host.ini and running ansible-playbook on all tests (or selected ones):
```
echo -e "[local]\nlocalhost ansible_python_interpreter=/usr/bin/python3 ansible_connection=local ansible_host=127.0.0.1" | tee host.ini
ansible-playbook -v --connection=local --inventory host.ini tests/tests_*.yml
```
this is what runs in the CI here.

richm:
re: tox - yeah, `uv` is the new hotness - the problem is that most of our stuff uses our custom `tox-lsr` tox plugin - using something like `uv` would require a pretty big rewrite - I do know that a lot of system roles developers struggle with tox-lsr for a wide variety of reasons - some things can be done in containers as Jakuje describes (`tox-lsr` has support for running tests in containers e.g. `tox -e container-ansible-core-2-21 -- --image-name centos-10 tests/tests_default.yml`) but for many system roles tests we need a VM - bootloader, kernel settings, selinux, virtual storage devices, some virtual network devices, reboot scenarios, etc. - so `tox -e qemu-ansible-core-2-21 -- ... tests/tests_something.yml` is the standard way to start a VM and run a test in it - (yeah, vagrant, I know)

If you are interested in getting `tox-lsr` to work, and you want a more interactive conversation, try #systemroles at libera.chat - I am `richm` there.

Zugschlus:
I think I got tox running for me now.

I have a strong argument towards accepting ansible lists, that way it is possible to set some options from different parts of the inventory (I frequently have global settings, part of them overridden from group variables, part of those again overridden by host variables), with being able to use ansible operations on a list instead of having to fiddle with strings that have different delimiters.

That way, you could for example have:
```
sshd_global:
  HostKeyAlgorithms:
    - "ssh-ed25519-cert-v01@openssh.com"
    - "ssh-ed25519"
```

and then add for systems that only have OpenSSH 7.4 that doesn't know ed25519 yet:

```
sshd_debian_buster:
  HostKeyAlgorithms:
    - "ssh-rsa2
```

and a local parent role could then merge the two mappings to ssh_config that is then fed to this role. This is significantly harder when the HostKeyAlgorithms are just a string listing the algorithms - you couldn't merge those strings because the parent role must for example take care to apply the correct number of commas in the correct order.

Zugschlus:
I took a plunge in having the make_option_lists script rewritten. Please take a look at https://github.com/Zugschlus/ansible-sshd/tree/zg/rewrite_make_option_lists and review make_option_lists.py. That script (126 lines) was generated by claude. 

This script is reasonably extendible just in case ssh upstream comes up with another scheme to write lists. 

As far as I can see, tox runs okay, the run ends with
```
  qemu-ansible-core-2-20: OK (52.28=setup[0.03]+cmd[52.26] seconds)
  congratulations :) (52.32 seconds)
```
I interpret this as "successful" and hope that's correct.

Once you indicate that this is worth going the way, I'm going to write tests and adapt the docs. But please tell me first that this is a viable option so that I don't waste my time.

Jakuje:
Looking at the compare:

https://github.com/willshersystems/ansible-sshd/compare/main...Zugschlus:ansible-sshd:zg/rewrite_make_option_lists

I think its a sensible approach. Just a few notes:
 * I would exclude the `comma/space` separator from the options/doc text as its really not used anywhere (aside of looking very odd)

@mattwillsher any thoughts?

Zugschlus:
I have now brought https://github.com/Zugschlus/ansible-sshd/tree/zg/rewrite_make_option_lists in a state almost ready to create a new PR. I will create the new PR once !386 has been sorted out (that is needed for the tests to pass).

I also didn't understand the part of the options/doc text that @Jakuje mentioned on Sep 21, please advise more detailed wha you want me to do.

I think this PR, !380 is ready to be closed, there will be a new one addressing the issues.

Jakuje:
> I also didn't understand the part of the options/doc text that @Jakuje mentioned on Sep 21, please advise more detailed wha you want me to do.

It was reference to the following line:

>     comma/space    -> list of lists; outer joined by ",", inner by " "

https://github.com/willshersystems/ansible-sshd/compare/main...Zugschlus:ansible-sshd:zg/rewrite_make_option_lists#diff-950cbef88c9a662c5f1abf4f89f762d8e29a3a4f438b3a1c408cb43ced4a9827R16

This kind of syntax is not used anywhere so I would not even define it in the script.

Zugschlus:
The next try is now https://github.com/willshersystems/ansible-sshd/pull/387, superseding this one. Should this stay open for the discussions or do you want it closed? Feel free to close.

